PR #12, opened 2017-07-05
WIP: Integration of Anderson acceleration of fixed-point iteration

devmotion:
This is a first draft trying to integrate Anderson acceleration. Unfortunately (currently) NLsolve accepts only vectors of real numbers, so I had to keep the simple Picard iteration for compatibility with Unitful.jl. Locally the tests passed already, however it is not benchmarked yet. I also tried to improve readability by adding some comments.

### DIFF
--- a/src/integrator_interface.jl
+++ b/src/integrator_interface.jl
@@ -1,67 +1,102 @@
 function savevalues!(integrator::DDEIntegrator,force_save=false)
+  # update ODE integrator with values of DDE integrator
   integrator.integrator.u = integrator.u
   integrator.integrator.k = integrator.k
   integrator.integrator.t = integrator.t
+
+  # add steps for interpolation when needed
   OrdinaryDiffEq.ode_addsteps!(integrator.integrator,integrator.f)
+
+  # update solution of ODE integrator
   savevalues!(integrator.integrator,force_save)
 end
 
 function postamble!(integrator::DDEIntegrator)
+  # update ODE integrator with values of DDE integrator
   integrator.integrator.u = integrator.u
   integrator.integrator.k = integrator.k
   integrator.integrator.t = integrator.t
+
+  # clean up solution of ODE integrator
   OrdinaryDiffEq.postamble!(integrator.integrator)
 end
 
 function perform_step!(integrator::DDEIntegrator)
   integrator.tprev = integrator.t # this is necessary to extrapolate from the current interval
+
+  # update ODE integrator with values of DDE integrator
   integrator.integrator.uprev = integrator.uprev
   integrator.integrator.tprev = integrator.tprev
   integrator.integrator.fsalfirst = integrator.fsalfirst
   integrator.integrator.t = integrator.t
   integrator.integrator.dt = integrator.dt
 
-  # if dt>max lag, then it's explicit so use Picard iteration
+  # if dt>min lag, then it's explicit so use fixed-point iteration
   if integrator.dt >minimum(integrator.prob.lags)
-
-    # the is done to correct the extrapolation
-    t_prev_cache = integrator.tprev
+    # this is done to correct the extrapolation
     t_cache = integrator.t
-    uprev_cache = integrator.uprev
-
-    numiters = 1
-    while true
-      if typeof(integrator.u) <: AbstractArray
-        copy!(integrator.u_cache,integrator.u)
-      else
-        integrator.u_cache = integrator.u
-      end
-      perform_step!(integrator,integrator.cache)
-
-      if typeof(integrator.resid) <: AbstractArray
-        integrator.resid .= (integrator.u .- integrator.u_cache)./(integrator.picardabstol .+ max.(abs.(integrator.u),abs.(integrator.u_cache)).*integrator.picardreltol)
-      else
-        integrator.resid = (integrator.u .- integrator.u_cache)./(integrator.picardabstol .+ max.(abs.(integrator.u),abs.(integrator.u_cache)).*integrator.picardreltol)
-      end
-
-      picardEEst = integrator.picardnorm(integrator.resid)
-      if picardEEst < 1 || numiters > integrator.max_picard_iters
-        break
-      end
-      if numiters == 1
-        integrator.integrator.tprev = integrator.t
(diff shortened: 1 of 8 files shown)


src/integrator_interface.jl
```diff
@@ -1,67 +1,102 @@
...
+          integrator.integrator.t = integrator.t+integrator.dt
+          integrator.integrator.uprev = integrator.u
+        end
+
+	# update ODE integrator with values of DDE integrator
+        integrator.integrator.u = integrator.u
+        integrator.integrator.k = integrator.k
+
+	# stop fixed-point iteration when residuals are small
+	integrator.picardnorm(integrator.resid) < 1 && break
```
devmotion: Currently the exit condition of the loop is checked directly after calculating the residuals. However, then no updates would be applied to the ODE integrator if the loop is left during the first iteration? For now, I moved it to the end - the same way Anderson accelerated fixed-point iteration with NLsolve does it right now.

src/integrator_interface.jl
```diff
@@ -1,67 +1,102 @@
...
+  # if dt>min lag, then it's explicit so use fixed-point iteration
   if integrator.dt >minimum(integrator.prob.lags)
-
-    # the is done to correct the extrapolation
-    t_prev_cache = integrator.tprev
+    # this is done to correct the extrapolation
     t_cache = integrator.t
-    uprev_cache = integrator.uprev
+    tprev_cache = integrator.tprev
+    if typeof(integrator.uprev_cache) <: AbstractArray
```
devmotion: I thought it might be useful to save allocations for `uprev_cache` and add a new field to the integrator.

src/integrator_type.jl
```diff
@@ -9,11 +9,13 @@ type DDEIntegrator{algType<:OrdinaryDiffEqAlgorithm,uType,tType,absType,relType,
...
-  picardabstol::absType
-  picardreltol::relType
+  uprev_cache::uType
+  fixedpoint_abstol::absType
+  fixedpoint_reltol::relType
   resid::residType # This would have to resize for resizing DDE to work
   picardnorm::NType
-  max_picard_iters::Int
+  max_fixedpoint_iters::Int
+  m::Int
```
devmotion: Maybe a more meaningful should be used instead of `m`?

src/integrator_interface.jl
ChrisRackauckas: That doesn't save allocations since `uprev_cache = integrator.uprev` is just saving the pointer to another variable.

ChrisRackauckas:
Would it be too much to separate out the naming changes + comments to a separate PR? Those are really good, and I would like those merged immediately. And separating those will make reviewing maintaining the change to Andersson much easier (it'll be much easier to find out if anything in the future breaks due to that change if it's isolated).

src/iteration_function.jl
```diff
@@ -0,0 +1,41 @@
+# nlsolve optimizes subtypes of AbstractDifferentiableMultivariateFunction
+mutable struct IterationFunction{I<:AbstractDDEIntegrator} <: AbstractDifferentiableMultivariateFunction
```
devmotion: "Regular" functions are always encapsulated `DifferentiableMultivariateFunction` by NLsolve. I'm not sure whether there is an advantage in defining a subtype of `AbstractDifferentiableMultivariateFunction` with the self-referential field `f!` instead of just subtyping `Function`

src/integrator_type.jl
ChrisRackauckas: Yes, I prefer descriptive variable names. `history_size`?

src/integrator_interface.jl
devmotion: Oh yes, my bad. So the question is rather: why is it saved in the first place? If entries of `integrator.uprev` are changed, entries of `uprev_cache` are also changed - so when we later set `integrator.uprev` to `uprev_cache` again, entries of `integrator.uprev` are in fact not updated to any prior values. Moreover, just by skimming through `low_order_rk_integrators.jl` in OrdinaryDiffEq.jl it seems that `integrator.uprev` is not updated by `perform_step!` - but I assume there are algorithms which also update `uprev`? Otherwise `uprev_cache` could be removed completely.

src/integrator_type.jl
devmotion: That sounds good 👍

devmotion:
I'll move the comments + naming changes to a seperate PR. Since I'm on a trip until the beginning of next week it might take some days, but hopefully I can do it tomorrow.

ChrisRackauckas:
Take your time. Have a nice trip. I think what's left here is resolving the conflicts and then the review.

devmotion:
I would like to complete https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/17 first. If it is merged it is probably easier to update this PR, and also the changes can be restricted to the implementation of Anderson acceleration only.

ChrisRackauckas:
I wonder if this is the right approach. NLsolve seems to allocate a bunch:

https://github.com/JuliaNLSolvers/NLsolve.jl/blob/master/src/anderson.jl#L17

and it'll do a dynamic dispatch each time too. This makes it fine for longer fixed point iteration problems, but for quick problems that's going to have a big overhead. 

Given that an the fact that most of the changes in this PR are due to commenting which was another PR, I think this can be closed. However, one approach to getting Anderson acceleration may be to rebase this PR, get it done, then test a new Anderson acceleration implementation against it.

devmotion:
It would be preferable to allocate these arrays once in the beginning of our algorithm, so they could be reused in every fixed-point iteration (although there is still a not in-place `circshift`...). So maybe we could just copy the Anderson algorithm and adapt it to our needs, without relying on NLSolve? I think our use case might be too special - otherwise one could try to add these changes to the implementation in NLSolve.

devmotion:
But I also guess we should close this PR for now.

ChrisRackauckas:
>So maybe we could just copy the Anderson algorithm and adapt it to our needs, without relying on NLSolve? I think our use case might be too special - otherwise one could try to add these changes to the implementation in NLSolve.

Yeah, I feel bad for asking for it in NLsolve but yes it seems we need a few specialized changes so that will make their implementation not work for our usage.

PR #13, opened 2017-07-07
Naming changes, comments, and consistent formatting

devmotion:
Hi!

I finally found the time to separate the naming changes and comments, together with some new additions, from https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/12. Moreover, I reformatted most parts of the code to follow the [Julia guidelines for code contributions](https://github.com/JuliaLang/julia/blob/master/CONTRIBUTING.md#general-formatting-guidelines-for-julia-code-contributions), since the same conventions, like e.g. line breaks after 92 characters, [should apply to docstrings](https://docs.julialang.org/en/stable/manual/documentation/) anyway.

Beside that, the only change to the code is the replacement of `type` with `struct` - or should these changes be moved to another PR?

I guess, the most comments are added to `integrator_interface.jl` and `solve.jl` - hopefully they can help for understanding the implementation.

### DIFF
--- a/src/integrator_interface.jl
+++ b/src/integrator_interface.jl
@@ -1,137 +1,217 @@
-function savevalues!(integrator::DDEIntegrator,force_save=false)
-  integrator.integrator.u = integrator.u
-  integrator.integrator.k = integrator.k
-  integrator.integrator.t = integrator.t
-  OrdinaryDiffEq.ode_addsteps!(integrator.integrator,integrator.f)
-  savevalues!(integrator.integrator,force_save)
+"""
+    savevalues!(integrator::DDEIntegrator, force_save=false)
+
+Update solution of `integrator`, if necessary or forced by `force_save`.
+"""
+function savevalues!(integrator::DDEIntegrator, force_save=false)
+    # update ODE integrator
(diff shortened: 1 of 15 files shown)


src/integrator_interface.jl
```diff
@@ -1,137 +1,217 @@
-function savevalues!(integrator::DDEIntegrator,force_save=false)
-  integrator.integrator.u = integrator.u
-  integrator.integrator.k = integrator.k
-  integrator.integrator.t = integrator.t
-  OrdinaryDiffEq.ode_addsteps!(integrator.integrator,integrator.f)
-  savevalues!(integrator.integrator,force_save)
+"""
+    savevalues!(integrator::DDEIntegrator, force_save=false)
+
+Update solution of `integrator`, if necessary or forced by `force_save`.
```
devmotion: According to the Julia documentation guide lines, "In general, only the most generic method should be documented" - so maybe this (and other docstrings in `integrator_interface.jl`) should be moved to OrdinaryDiffEq.jl rather?

ChrisRackauckas: I think it's fine to have it both places. I'll accept it here.

PR #14, opened 2017-07-11
Use DiffEqProblemLibrary and add tests

devmotion:
Hi!

Already shortly after https://github.com/JuliaDiffEq/DiffEqProblemLibrary.jl/pull/9 was merged I wanted to open this PR in order to use `DiffEqProblemLibrary.jl` in the tests. However, by adding some additional tests I discovered that the unconstrained algorithm computes different solutions for in-place and not in-place functions. I tried to fix this problem, but without success yet. I'll open an issue regarding this problem, the failing tests are commented out in this PR.

### DIFF
--- a/test/REQUIRE
+++ b/test/REQUIRE
@@ -1,3 +1,4 @@
 DiffEqDevTools
 DiffEqCallbacks
+DiffEqProblemLibrary
 Unitful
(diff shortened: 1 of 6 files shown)


ChrisRackauckas:
Instead of commenting them out, can you add `@test_broken`? Need to check around for a place where the values alias. Thanks for catching it.

devmotion:
Tests work locally, but will fail on Travis etc because there is no release of `DiffEqProblemLibrary.jl` with DDE problems yet.

ChrisRackauckas:
Got it. I'll just tag that right now.

ChrisRackauckas:
https://github.com/JuliaLang/METADATA.jl/pull/10216

Issue #15, opened 2017-07-11
Different solutions for in-place and not in-place functions with unconstrained algorithm


devmotion:
Hi!

As already mentioned in https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/14, the unconstrained algorithm computes different solutions for problems with in-place and not in-place functions. A small example:

```julia
julia> using DelayDiffEq, DiffEqBase, OrdinaryDiffEq, DiffEqProblemLibrary
julia> alg = MethodOfSteps(BS3(); constrained=false)
julia> u₀ = 1.0
julia> prob = prob_dde_1delay_scalar_notinplace(u₀)
julia> sol = solve(prob, alg)
julia> prob2 = prob_dde_1delay(u₀)
julia> sol2 = solve(prob2, alg)
julia> sol.t == sol2.t
false

julia> sol.u == sol2[1, :]
false

julia> sol.errors
Dict{Symbol,Float64} with 3 entries:
  :l∞    => 3.33289e-5
  :final => 1.86351e-5
  :l2    => 1.40652e-5

julia> sol2.errors
Dict{Symbol,Float64} with 3 entries:
  :l∞    => 3.31446e-5
  :final => 1.86005e-5
  :l2    => 1.401e-5
```
I assume the problem originates from the DDE `integrator` and the ODE `integrator.integrator` pointing, among others, both to the same arrays `u`, `uprev`, and `k` in the in-place case. So for some assignments such as https://github.com/JuliaDiffEq/DelayDiffEq.jl/blob/master/src/integrator_interface.jl#L96, which is executed in the first iteration, the value is only updated once for not in-place integrators but immediately in every step for in-place integrators. Moreover, even during calculations in https://github.com/JuliaDiffEq/DelayDiffEq.jl/blob/master/src/integrator_interface.jl#L69 any change of `u` or `k` is immediately propagated to `integrator.integrator`, and thus might also change the interpolation which is not desired, I guess. I was able to reduce the difference of the errors in the example above by completely separating `integrator.u` and `integrator.k` also for in-place functions, i.e. by creating a copy of both arrays during initialization and always using `recursivecopy!` instead of assignments such as `integrator.integrator.u = integrator.u`, but I got still a tiny difference. 

[...]

Issue #16, opened 2017-07-13
Methods with lazy interpolants will not work with MethodOfSteps


ChrisRackauckas:
This means the Verner methods are excluded. It likely has to do something with the adding of the extra steps not happening correctly. It probably needs https://github.com/JuliaDiffEq/DelayDiffEq.jl/issues/15#issuecomment-314757874 .

PR #17, opened 2017-07-13
WIP: Fix different solutions

devmotion:
I finally found the missing piece, and could fix https://github.com/JuliaDiffEq/DelayDiffEq.jl/issues/15 for BS3 algorithms. The fix can easily be extended to all other affected algorithms, however, it might be preferable to add it to `OrdinaryDiffEq.jl`.

Besides the already discussed separation of `u` and `k` in `perform_steps!` it is necessary to update the value of `fsalfirst` in `k` (if existent): for in-place functions this happens automatically via pointers, when `fsalfirst` is updated in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/integrators/integrator_utils.jl#L280; for not in-place functions `fsalfirst` is updated in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/integrators/integrator_utils.jl#L282, but this change is propagated to `k` only at the end of the next execution of `perform_step!` in https://github.com/devmotion/OrdinaryDiffEq.jl/blob/master/src/integrators/low_order_rk_integrators.jl#L43. Hence the interpolation, which depends on `k`, differs. This should apply to all not in-place algorithms which save `fsalfirst` in `k` and use its entry in `k` for interpolation; I could provoke the problem with Tsit5 and apply the same fix. Only the fix for BS3 is included here so far, since I was not sure whether a fix in OrdinaryDiffEq.jl would be more appropriate.

The tests pass locally but need https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/80.

### DIFF
--- a/src/integrator_interface.jl
+++ b/src/integrator_interface.jl
@@ -55,7 +55,7 @@ function perform_step!(integrator::DDEIntegrator)
         # update cached error estimate of integrator
         integrator.integrator.EEst = integrator.EEst
 
-        # save value u(tprev) and interpolation data of interval [tprev, t] of ODE
+        # save dt, u(tprev) and interpolation data of interval [tprev, t] of ODE
         # integrator since they are overwritten by fixed-point iteration
         if typeof(integrator.uprev_cache) <: AbstractArray
             recursivecopy!(integrator.uprev_cache, integrator.integrator.uprev)
@@ -63,6 +63,7 @@ function perform_step!(integrator::DDEIntegrator)
             integrator.uprev_cache = integrator.integrator.uprev
         end
         recursivecopy!(integrator.k_integrator_cache, integrator.integrator.k)
+        integrator.integrator.dtcache = integrator.integrator.dt
 
         # move ODE integrator to interval [t, t+dt] to use interpolation of ODE integrator
         # in the next iterations when evaluating the history function
@@ -139,7 +140,7 @@ function perform_step!(integrator::DDEIntegrator)
         # u(tprev) and u(t), and interpolation data k of this interval
         integrator.integrator.t = integrator.t
         integrator.integrator.tprev = integrator.tprev
-        integrator.integrator.dt = integrator.integrator.t - integrator.integrator.tprev
+        integrator.integrator.dt = integrator.integrator.dtcache
         if typeof(integrator.u) <: AbstractArray
             recursivecopy!(integrator.integrator.u, integrator.uprev)
             recursivecopy!(integrator.integrator.uprev, integrator.uprev_cache)
--- a/src/integrator_interface.jl
+++ b/src/integrator_interface.jl
@@ -4,10 +4,23 @@
 Update solution of `integrator`, if necessary or forced by `force_save`.
 """
 function savevalues!(integrator::DDEIntegrator, force_save=false)
-    # update ODE integrator
-    integrator.integrator.u = integrator.u
-    integrator.integrator.k = integrator.k
+    # update ODE integrator to interval [tprev, t] with corresponding values
+    # u(tprev) and u(t), and interpolation data k of this interval
     integrator.integrator.t = integrator.t
+    integrator.integrator.tprev = integrator.tprev
+
+    # copy u(tprev) since it is overwritten by integrator at the end of apply_step!
+    if typeof(integrator.u) <: AbstractArray
+        recursivecopy!(integrator.integrator.u, integrator.u)
+        recursivecopy!(integrator.integrator.uprev, integrator.uprev)
+    else
+        integrator.integrator.u = integrator.u
+        integrator.integrator.uprev = integrator.uprev
+    end
+
+    # copy interpolation data (fsalfirst overwritten at the end of apply_step!, which also
+    # updates k[1] when using chaches for which k[1] points to fsalfirst)
+    recursivecopy!(integrator.integrator.k, integrator.k)
 
     # add steps for interpolation to ODE integrator when needed
     OrdinaryDiffEq.ode_addsteps!(integrator.integrator, integrator.f)
@@ -23,9 +36,16 @@ Clean up solution of `integrator`.
 """
 function postamble!(integrator::DDEIntegrator)
     # update ODE integrator
-    integrator.integrator.u = integrator.u
-    integrator.integrator.k = integrator.k
     integrator.integrator.t = integrator.t
+    if typeof(integrator.u) <: AbstractArray
+        recursivecopy!(integrator.integrator.u, integrator.u)
+    else
+        integrator.integrator.u = integrator.u
+    end
+    recursivecopy!(integrator.integrator.k, integrator.k)
+
+    # add steps for interpolation to ODE integrator when needed
+    OrdinaryDiffEq.ode_addsteps!(integrator.integrator, integrator.f)
 
     # clean up solution of ODE integrator
     OrdinaryDiffEq.postamble!(integrator.integrator)
@@ -37,86 +57,131 @@ end
 Calculate next step of `integrator`.
 """
 function perform_step!(integrator::DDEIntegrator)
-    # update ODE integrator
-    integrator.integrator.uprev = integrator.uprev
-    integrator.integrator.tprev = integrator.t # to extrapolate from current interval
-    integrator.integrator.t = integrator.t
-    integrator.integrator.dt = integrator.dt
+    # cache error estimate of integrator and interpolation data of interval [tprev, t]
+    # (maybe with already updated entry k[1] = fsalfirst == fsallast, if k[1] points to
+    # fsalfirst) to be able to reset the corresponding variables in case calculation results
+    # in numbers that are not finite
+    recursivecopy!(integrator.k_cache, integrator.k)
+    integrator.integrator.EEst = integrator.EEst
 
-    # separate u to avoid influence of calculations on interpolation and to calculate
-    # residuals
-    if typeof(integrator.u) <: AbstractArray
-        recursivecopy!(integrator.u_cache, integrator.u)
-    else
-        integrator.u_cache = integrator.u
-    end
-    integrator.integrator.u = integrator.u_cache
+    # add steps to interpolation data ODE integrator if necessary
+    OrdinaryDiffEq.ode_addsteps!(integrator.integrator, integrator.f)
 
-    # separate k to avoid influence of calculations on interpolation
-    # constant caches that contain fsalfirst must be updated for interpolation
-    # has to be changed all such caches - move to OrdinaryDiffEq?
-    if typeof(integrator.cache) <: Union{BS3ConstantCache}
-        integrator.k[1] = integrator.fsalfirst
-    end
-    integrator.integrator.k = recursivecopy(integrator.k)
+    # perform always at least one calculation
+    perform_step!(integrator, integrator.cache)
 
-    # if dt is greater than the minimal lag, then it's explicit so use fixed-point iteration
-    if integrator.dt > minimum(integrator.prob.lags)
-        numiters = 1
+    # shrink interpolation data of ODE problem
+    resize!(integrator.integrator.k, integrator.integrator.kshortsize)
+
+    # if dt is greater than the minimal lag, then use a fixed-point iteration
+    if integrator.dt > minimum(integrator.prob.lags) && isfinite(integrator.EEst)
+
+        # update cached error estimate of integrator
+        integrator.integrator.EEst = integrator.EEst
+
+        # save value u(tprev) and interpolation data of interval [tprev, t] of ODE
+        # integrator since they are overwritten by fixed-point iteration
+        if typeof(integrator.uprev_cache) <: AbstractArray
+            recursivecopy!(integrator.uprev_cache, integrator.integrator.uprev)
+        else
+            integrator.uprev_cache = integrator.integrator.uprev
+        end
+        recursivecopy!(integrator.k_integrator_cache, integrator.integrator.k)
+
+        # move ODE integrator to interval [t, t+dt] to use interpolation of ODE integrator
+        # in the next iterations when evaluating the history function
+        integrator.integrator.t = integrator.t + integrator.dt
+        integrator.integrator.tprev = integrator.t
+        if typeof(integrator.integrator.uprev) <: AbstractArray
+            recursivecopy!(integrator.integrator.uprev, integrator.uprev)
+        else
+            integrator.integrator.uprev = integrator.uprev
+        end
+
+        numiters=1
 
         while true
+            # update value u(t+dt) and interpolation data of interval [t, t+dt] that are
+            # used for the interpolation of the history function in the next iteration
+            if typeof(integrator.u) <: AbstractArray
+                recursivecopy!(integrator.integrator.u, integrator.u)
+            else
+                integrator.integrator.u = integrator.u
+            end
+            recursivecopy!(integrator.integrator.k, integrator.k)
+
+            # add steps to interpolation data of ODE integrator if necessary
+            OrdinaryDiffEq.ode_addsteps!(integrator.integrator, integrator.f)
+
             # calculate next step
             perform_step!(integrator, integrator.cache)
 
-            # calculate residuals
+            # shrink interpolation data of ODE integrator
+            resize!(integrator.integrator.k, integrator.integrator.kshortsize)
+
+            # calculate residuals of fixed-point iteration
             if typeof(integrator.resid) <: AbstractArray
-                @. integrator.resid = (integrator.u - integrator.u_cache) /
+                @. integrator.resid = (integrator.u - integrator.integrator.u) /
                     @muladd(integrator.fixedpoint_abstol + max(abs(integrator.u),
-                                                               abs(integrator.u_cache)) *
+                                                               abs(integrator.integrator.u)) *
                             integrator.fixedpoint_reltol)
             else
-                integrator.resid = @. (integrator.u - integrator.u_cache) /
+                integrator.resid = @. (integrator.u - integrator.integrator.u) /
                     @muladd(integrator.fixedpoint_abstol + max(abs(integrator.u),
-                                                               abs(integrator.u_cache)) *
+                                                               abs(integrator.integrator.u)) *
                             integrator.fixedpoint_reltol)
             end
-
-            # stop fixed-point iteration when residuals are small or maximal number of steps
-            # is exceeded
             fixedpointEEst = integrator.fixedpoint_norm(integrator.resid)
-            if fixedpointEEst < 1 || numiters > integrator.max_fixedpoint_iters
+
+            # stop fixed-point iteration when error estimate of integrator or error estimate
+            # of fixed-point iteration are not finite
+            if !(isfinite(fixedpointEEst)) || !(isfinite(integrator.EEst))
+                # assure that integrator is reset to cached values
+                integrator.EEst = max(fixedpointEEst, integrator.EEst)
                 break
             end
 
-            # special update of ODE integrator after the first iteration step
-            # to use interpolation of ODE integrator in the next iterations
-            # when evaluating the history function
-            if numiters == 1
-                integrator.integrator.t = integrator.t + integrator.dt
-            end
+            # update cached value of error estimate of integrator with a combined error
+            # estimate of both integrator and fixed-point iteration
+            # this prevents acceptance of steps with poor performance in fixed-point
+            # iteration
+            integrator.integrator.EEst = max(fixedpointEEst, integrator.EEst)
 
-            # update cached values of u and k
-            if typeof(integrator.u) <: AbstractArray
-                recursivecopy!(integrator.u_cache, integrator.u)
-            else
-                integrator.u_cache = integrator.u
+            # stop fixed-point iteration when error estimate is small or maximal number of
+            # steps is exceeded
+            if integrator.integrator.EEst <= 1 || numiters > integrator.max_fixedpoint_iters
+                # update error estimate with combined error estimate
+                integrator.EEst = integrator.integrator.EEst
+                break
             end
-            integrator.integrator.u = integrator.u_cache
-
-            recursivecopy!(integrator.integrator.k, integrator.k)
 
             numiters += 1
         end
 
-        # reset time of ODE integrator
+        # reset ODE integrator to interval [tprev, t] with corresponding values
+        # u(tprev) and u(t), and interpolation data k of this interval
         integrator.integrator.t = integrator.t
-    else # no iterations
-        perform_step!(integrator, integrator.cache)
+        integrator.integrator.tprev = integrator.tprev
+        if typeof(integrator.u) <: AbstractArray
+            recursivecopy!(integrator.integrator.u, integrator.uprev)
+            recursivecopy!(integrator.integrator.uprev, integrator.uprev_cache)
+        else
+            integrator.integrator.u = integrator.uprev
+            integrator.integrator.uprev = integrator.uprev_cache
+        end
+        recursivecopy!(integrator.integrator.k, integrator.k_integrator_cache)
     end
 
-    # update ODE integrator
-    integrator.integrator.u = integrator.u
-    integrator.integrator.k = integrator.k
+    # if error estimate of integrator is not a finite number reset it to last cached error
+    # estimate or 2, and reset interpolation data of integrator to interpolation data of
+    # interval [tprev, t] (maybe with updated entry k[1]) before calculation of current step
+    # then current step will not be accepted, time step dt will be decreased,
+    # and calculation of next step will be repeated starting with the same
+    # initial interpolation data
+    if !isfinite(integrator.EEst)
+        integrator.EEst = max(2, integrator.integrator.EEst) # EEst must be > 1
+        recursivecopy!(integrator.k, integrator.k_cache)
+    end
 end
 
 """
@@ -127,8 +192,14 @@ Set initial values of `integrator`.
 function initialize!(integrator::DDEIntegrator)
     initialize!(integrator, integrator.cache, integrator.f)
 
-    # set also initial values of ODE integrator
-    initialize!(integrator.integrator, integrator.cache, integrator.f)
+    # interpolation data of integrator and ODE integrator have to be cached
+    # when next step is calculated
+    integrator.k_cache = recursivecopy(integrator.k)
+    integrator.k_integrator_cache = recursivecopy(integrator.k)
+
+    # copy interpolation data to ODE integrator
+    integrator.integrator.kshortsize = integrator.kshortsize
+    integrator.integrator.k = recursivecopy(integrator.k)
 end
 
 """
--- a/src/integrator_interface.jl
+++ b/src/integrator_interface.jl
@@ -37,34 +37,34 @@ end
 Calculate next step of `integrator`.
 """
 function perform_step!(integrator::DDEIntegrator)
-    # update previous time to extrapolate from current interval
-    integrator.tprev = integrator.t
-
     # update ODE integrator
     integrator.integrator.uprev = integrator.uprev
-    integrator.integrator.tprev = integrator.tprev
-    integrator.integrator.fsalfirst = integrator.fsalfirst
+    integrator.integrator.tprev = integrator.t # to extrapolate from current interval
     integrator.integrator.t = integrator.t
     integrator.integrator.dt = integrator.dt
 
+    # separate u to avoid influence of calculations on interpolation and to calculate
+    # residuals
+    if typeof(integrator.u) <: AbstractArray
+        recursivecopy!(integrator.u_cache, integrator.u)
+    else
+        integrator.u_cache = integrator.u
+    end
+    integrator.integrator.u = integrator.u_cache
+
+    # separate k to avoid influence of calculations on interpolation
+    # constant caches that contain fsalfirst must be updated for interpolation
+    # has to be changed all such caches - move to OrdinaryDiffEq?
+    if typeof(integrator.cache) <: Union{BS3ConstantCache}
+        integrator.k[1] = integrator.fsalfirst
+    end
+    integrator.integrator.k = recursivecopy(integrator.k)
+
     # if dt is greater than the minimal lag, then it's explicit so use fixed-point iteration
     if integrator.dt > minimum(integrator.prob.lags)
-
-        # save these values to correct the extrapolation after the last iteration
-        tprev_cache = integrator.tprev
-        t_cache = integrator.t # same as tprev_cache?
-        uprev_cache = integrator.uprev
-
         numiters = 1
+
         while true
-
-            # save u to calculate residuals (u is overwritten in calculation of next step)
-            if typeof(integrator.u) <: AbstractArray
-                copy!(integrator.u_cache, integrator.u)
-            else
-                integrator.u_cache = integrator.u
-            end
-
             # calculate next step
             perform_step!(integrator, integrator.cache)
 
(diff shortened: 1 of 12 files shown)


src/integrator_interface.jl
```diff
@@ -37,34 +37,34 @@ end
...
+    integrator.integrator.u = integrator.u_cache
 
-        # save these values to correct the extrapolation after the last iteration
-        tprev_cache = integrator.tprev
-        t_cache = integrator.t # same as tprev_cache?
-        uprev_cache = integrator.uprev
+    # separate k to avoid influence of calculations on interpolation
+    # constant caches that contain fsalfirst must be updated for interpolation
+    # has to be changed all such caches - move to OrdinaryDiffEq?
+    if typeof(integrator.cache) <: Union{BS3ConstantCache}
```
devmotion: Just forgot to remove `Union` after experimenting with Tsit5 😄

ChrisRackauckas: what's this part for?

devmotion: To update `k[1]`. For some caches like `BS3ConstantCache` it contains `fsalfirst` - however, at this point of the algorithm its an old, outdated value of `fsalfirst`, and hence also any interpolation that uses `k[1]` is wrong. There are no problems with `BS3Cache` since it only points to `fsalfirst` and hence the value of `k[1]` always equals `fsalfirst`.

Of course, one has to add all affected caches (there are many more besides `BS3ConstantCache` - all not-inplace caches that contain `fsalfirst` and use it for interpolation) to this `Union`. I just wanted to wait til I know whether you would like to apply this update of `k` here or rather in `OrdinaryDiffEq.jl` when `fsalfirst` is updated.

ChrisRackauckas: won't this make `k[1] == k[2]` in the first round, making the interpolation/extrapolation invalid?

devmotion: Yes, it will make `k[1] == k[2]` in the first round. However, I'm still not completely sure about which values one would like to use for the interpolation - I mean, I know what we want to achieve but the problem is the correct assignment of `u`, `uprev`, `t`, and `tprev` for `integrator.integrator`. With the current (and the old) setup we have `u == uprev` (by the ODE algorithm) and `t == tprev` (by our assignments) in the first round. In the next rounds `u` is always updated to the newest estimate, `uprev` stays fixed, `t` is fixed to `integrator.t + integrator.dt`, i.e. the time point after this step, and `tprev` is fixed to `integrator.t`, i.e. the current time point.

ChrisRackauckas: it should never have `k[1] == k[2]`. That would make it a constant extrapolation instead of the Hermite. In the first round it should stay in the previous setup, unchanged. After the first iteration, `tprev = integrator.t`, `t = integrator.t + integrator.dt`, `uprev` should be `u(tprev)`, and `u` should be `u(t)`

src/integrator_interface.jl
```diff
@@ -37,34 +37,34 @@ end
...
 """
 function perform_step!(integrator::DDEIntegrator)
-    # update previous time to extrapolate from current interval
-    integrator.tprev = integrator.t
-
     # update ODE integrator
     integrator.integrator.uprev = integrator.uprev
-    integrator.integrator.tprev = integrator.tprev
-    integrator.integrator.fsalfirst = integrator.fsalfirst
+    integrator.integrator.tprev = integrator.t # to extrapolate from current interval
```
ChrisRackauckas: This can't update yet: it doesn't have valid values to extrapolate with.

src/integrator_interface.jl
```diff
@@ -81,44 +81,42 @@ function perform_step!(integrator::DDEIntegrator)
...
         perform_step!(integrator, integrator.cache)
     end
 
-    #integrator.u = integrator.integrator.u
-    #integrator.fsallast = integrator.integrator.fsallast
-    #if integrator.opts.adaptive
-    #  integrator.EEst = integrator.integrator.EEst
-    #end
+    # update ODE integrator
+    integrator.integrator.u = integrator.u
```
ChrisRackauckas: what's this for?

src/integrator_interface.jl
devmotion: Ah OK, then it is also incorrect on the `master` branch, since it just shortens lines 41 and 45 of the current implementation. What values exactly are missing? The value at time `integrator.t` is `integrator.uprev = integrator.u` at this point.

ChrisRackauckas: Oh wait, was `integrator.integrator.tprev` never updated in the previous round? I think I'm going to have to run through it all again and print values out because it's definitely tricky.

src/integrator_interface.jl
devmotion: That's exactly one of the reasons why I was confused about this part. So the problem in our case is rather that for in-place algorithms with `k[1] = integrator.fsalfirst` and `k[2] = integrator.fsallast` (like BS3Cache) we always automatically have `k[1] == k[2]` at this point.... Maybe it's best to always separate `k` by `integrator.integrator.k = recursivecopy(integrator.k)` right after initialization and then use `recursivecopy!` in all other places? But I guess I have to try it to see whether it actually works...

src/integrator_interface.jl
devmotion: To update the ODE integrator. Otherwise `integrator.integrator.u` still contains `u(integrator.t)`, when peforming a step without iterations, or `u(integrator.t+integrator.dt)` after the second to last iteration, when performing fixed-point iterations. For any interpolations that occur after this step it should rather be the current, updated value `u(integrator.t+integrator.dt)`.

devmotion:
Since both the current implementation and the first draft of this PR are not completely correct, I try to summarize both what we want to achieve and my suggestions/ideas, even though it is quite lengthy. I want to provide a rough overview of our desired algorithm using an example, hopefully as an assistance and a reminder to myself. Maybe it can also speed up the process of fixing the discovered problems. I surely made some mistakes, so please correct anything that is wrong.

## Example

Assume we already calculated a solution *u(t)* up to time point *t<sub>1</sub>*, and the last step of our computation we moved from time point *t<sub>0</sub>* to *t<sub>1</sub>* such that *[t<sub>0</sub>, t<sub>1</sub>]* is  the last calculated interval. Let *dt* be the proposed next time step, i.e. we would like to advance to time point *t<sub>1</sub> + dt* in the next step. Moreover, let *k<sub>[a, b]</sub>(x, y)* be the interpolation data of time interval *[a, b]* with function values *x* and *y* at *a* and *b*, respectively. Then our algorithm works in the following fashion:

* `DelayDiffEq.perform_steps!`: When stepping into that method a subset of the variable assignments is given by
  
  Variable | Value
  -----------|--------
  integrator.u | usually *u(t<sub>1</sub>)* (see notes)
  integrator.uprev | *u(t<sub>1</sub>)*
  integrator.t | *t<sub>1</sub>*
  integrator.tprev | *t<sub>0</sub>*
  integrator.k | usually *k<sub>[t<sub>0</sub>, t<sub>1</sub>]</sub>(u(t<sub>0</sub>), u(t<sub>1</sub>))* (see notes)
  integrator.dt | *dt*
  integrator.integrator.u | *u(t<sub>1</sub>)*
  integrator.integrator.uprev | *u(t<sub>0</sub>)*
  integrator.integrator.t | *t<sub>1</sub>*
  integrator.integrator.tprev | *t<sub>0</sub>*
  integrator.integrator.k | *k<sub>[t<sub>0</sub>, t<sub>1</sub>]</sub>(u(t<sub>0</sub>), u(t<sub>1</sub>))*

  Notes:
    * FSAL algorithms for in-place function already updated `integrator.fsalfirst` to `integrator.fsallast` in `OrdinaryDiffEq.apply_step!`, and hence if `integrator.k[1]` is set up to point to `integrator.fsalfirst` it holds `integrator.k[1] == integrator.fsalfirst == integrator.fsallast` at this point. Hence in that case `integrator.k` DOES NOT equal *k<sub>[t<sub>0</sub>, t<sub>1</sub>]</sub>(u(t<sub>0</sub>), u(t<sub>1</sub>))*, and in particular `integrator.integrator.k != integrator.k`.
  * If we already calculated a value for *u(t<sub>1</sub>+T)* for some *T > dt* in the last iteration which was not accepted then `integrator.u` equals that value instead.
  
  We then calculate *u(t<sub>1</sub>+dt)*:
    1. If *dt* is less or equal to the minimal lag of our DDE system we just apply the given ODE algorithm. Out of the variables above, only `integrator.u` and `integrator.k` are updated to *u(t<sub>1</sub>+dt)* and *k<sub>[t<sub>1</sub>, t<sub>1</sub> + dt]</sub>(u(t<sub>1</sub>), u(t<sub>1</sub> + dt))*, respectively.

    2. In the other case we approximate *u(t<sub>1</sub> + dt)* by a fixed-point iteration:
       * In the first iteration step we just apply the given ODE algorithm and obtain a first approximation *u<sup>1</sup>(t<sub>1</sub>+dt)*.
       * For the next iteration steps we fix `integrator.integrator.t`, `integrator.integrator.tprev`, and `integrator.integrator.uprev` to *t<sub>1</sub> + dt*, *t<sub>1</sub>*, and *u(t<sub>1</sub>)*, respectively. Moreover, after the *n*-th iteration step we update `integrator.integrator.u` to *u<sup>n</sup>(t<sub>1</sub> + dt)* and `integrator.integrator.k` to *k<sub>[t<sub>1</sub>, t<sub>1</sub> + dt]</sub>(u(t<sub>1</sub>), u<sup>n</sup>(t<sub>1</sub> + dt))*. The fixed-point iteration is stopped if the maximal number of iterations is reached or the error estimate of the fixed point iteration, given by *||du / (abstol + reltol × U)||* where *du = u<sup>n</sup>(t<sub>1</sub> + dt) - u<sup>n-1</sup>(t<sub>1</sub> + dt)* and *U = max(|u<sup>n</sup>(t<sub>1</sub> + dt)|, |u<sup>n-1</sup>(t<sub>1</sub> + dt)|)*, is below 1.
    
       When the fixed-point iteration is stopped after the *m*-th iteration the last approximation *u<sup>m</sup>(t<sub>1</sub> + dt)* and the corresponding interpolation data *k<sub>[t<sub>1</sub>, t<sub>1</sub> + dt]</sub>(u(t<sub>1</sub>), u<sup>m</sup>(t<sub>1</sub> + dt))* are assigned to `integrator.u` and `integrator.k`. Moreover, we reset `integrator.integrator.u`, `integrator.integrator.uprev`, `integrator.integrator.t`, `integrator.integrator.tprev`, and `integrator.integrator.k` to the values before values before the fixed-point iteration, i.e. to *u(t<sub>1</sub>)*, *u(t<sub>0</sub>)*, *t<sub>1</sub>*, *t<sub>0</sub>*, and *k<sub>[t<sub>0</sub>, t<sub>1</sub>]</sub>(u(t<sub>0</sub>), u(t<sub>1</sub>))*.

* `OrdinaryDiffEq.loopfooter!`: Immediately after the calculation of *u(t<sub>1</sub> + dt)* this method is executed, and hence the variable assignments look similar as above:

  Variable | Value
  -----------|--------
  integrator.u | *u(t<sub>1</sub> + dt)*
  integrator.uprev | *u(t<sub>1</sub>)*
  integrator.t | *t<sub>1</sub>*
  integrator.tprev | *t<sub>0</sub>*
  integrator.k | *k<sub>[t<sub>1</sub>, t<sub>1</sub> + dt]</sub>(u(t<sub>1</sub>), u(t<sub>1</sub> + dt))*
  integrator.dt | *dt*
  integrator.integrator.u | *u(t<sub>1</sub>)*
  integrator.integrator.uprev | *u(t<sub>0</sub>)*
  integrator.integrator.t | *t<sub>1</sub>*
  integrator.integrator.tprev | *t<sub>0</sub>*
  integrator.integrator.k | *k<sub>[t<sub>0</sub>, t<sub>1</sub>]</sub>(u(t<sub>0</sub>), u(t<sub>1</sub>))*

  There are two cases:
    * If the step to time *t<sub>1</sub> + dt* and value *u(t<sub>1</sub> + dt)* is not accepted nothing changes.
    * If the step is accepted, `integrator.t` and `integrator.tprev` are updated to *t<sub>1</sub> + dt* and *t<sub>1</sub>*. Moreover, the next step size *dt<sub>1</sub>* is proposed and continuous and discrete callbacks are handled. Before, in between, or after callbacks are applied the method

         * `DelayDiffEq.savevalues!` is executed. It propagates changes to `integrator` introduced by the callbacks or just values of the before calculated time interval *[t<sub>1</sub>, t<sub>1</sub> + dt]* to `integrator.integrator`, i.e. we update `integrator.integrator.u` to `integrator.u`, `integrator.integrator.uprev` to `integrator.uprev`, `integrator.integrator.t` to `integrator.t`, `integrator.integrator.tprev` to `integrator.tprev`, and `integrator.integrator.k` to `integrator.k`. Hence without callbacks `integrator.integrator` is just upated to the time interval *[t<sub>1</sub>, t<sub>1</sub> + dt]*. The current values and the current interpolation of `integrator.integrator` are then used by 
         * `OrdinaryDiffEq.savevalues!` to save values at time points before `integrator.integrator.t`, i.e. *t<sub>1</sub> + dt* in the case without callbacks, to the solution `integrator.integrator.sol` (and implicitly also to `integrator.sol` since they both point to the same arrays). Moreover, at the end the interpolation data `integrator.integrator.k` is reduced to its minimal size `integrator.integrator.kshortsize` which depends on the chosen algorithm.

      If `integrator.u`, i.e. *u(t<sub>1</sub> + dt)*, is modified by a callback, `DelayDiffEq.handle_callback_modifiers!` marks `integrator.fsalfirst` for recalculation at the end of this step and updates the discontinuity tree of our system.

Now, if there are still stopping time points left but there is no stopping time point between *t<sub>1</sub>* and `integrator.t`, i.e. *t<sub>1</sub> + dt* without callbacks, we move on to
  
* `OrdinaryDiffEq.loopheader!`:

  * If the step to *t<sub>1</sub> + dt* was not accepted, the assignments noted in the table above are still in use. According to some formulas, depending on whether the calculated value *u(t<sub>1</sub> + dt)* was out of the domain, and accounting for given maximal or minimal step sizes and stopping time points, `integrator.dt` is updated with a reduced time step *ds < dt*. We then continue at the top in `DelayDiffEq.perform_step!` to find a better calculation *u(t<sub>1</sub> + ds)* using the variable assignment

  Variable | Value
  -----------|--------
  integrator.u | *u(t<sub>1</sub> + dt)*
  integrator.uprev | *u(t<sub>1</sub>)*
  integrator.t | *t<sub>1</sub>*
  integrator.tprev | *t<sub>0</sub>*
  integrator.k | *k<sub>[t<sub>1</sub>, t<sub>1</sub> + dt]</sub>(u(t<sub>1</sub>), u(t<sub>1</sub> + dt))*
  integrator.dt | *ds*
  integrator.integrator.u | *u(t<sub>1</sub>)*
  integrator.integrator.uprev | *u(t<sub>0</sub>)*
  integrator.integrator.t | *t<sub>1</sub>*
  integrator.integrator.tprev | *t<sub>0</sub>*
  integrator.integrator.k | *k<sub>[t<sub>0</sub>, t<sub>1</sub>]</sub>(u(t<sub>0</sub>), u(t<sub>1</sub>))*

  * If the step to *t<sub>1</sub> + dt* was accepted we go to

     * `OrdinaryDiffEq.apply_step!`: There `integrator.uprev`, which currently holds the value *u(t<sub>1</sub>)*, is updated to the value of `integrator.u`, i.e. *u(t<sub>1</sub> + dt)* without callbacks. Moreover, `integrator.dt` is changed from *dt* to *dt<sub>1</sub>*, i.e. we try to move to time point *t<sub>1</sub> + dt + dt<sub>1</sub>* in the next step. For FSAL algorithms the value of `integrator.fsallast` is copied to `integrator.fsalfirst` (so we can move on from *u(t<sub>1</sub> + dt)*) or, if we are at a discontinuity or a callback changed the value of *u(t<sub>1</sub> + dt)*, `integrator.fsalfirst` is recalculated.

     * Similar to the case of not accepted steps, the given maximal or minimal step sizes and stopping time points are also taken into account, and `integrator.dt` might be updated with a reduced time step *dt<sub>2</sub> < dt<sub>1</sub>*.

  Then, assuming no callbacks, the considered subset of variables is given by

  Variable | Value
  -----------|--------
  integrator.u | *u(t<sub>1</sub> + dt)*
  integrator.uprev | *u(t<sub>1</sub> + dt)*
  integrator.t | *t<sub>1</sub> + dt*
  integrator.tprev | *t<sub>1</sub>*
  integrator.k | usually *k<sub>[t<sub>1</sub>, t<sub>1</sub> + dt]</sub>(u(t<sub>1</sub>), u(t<sub>1</sub> + dt))* (see notes above)
  integrator.dt | *dt<sub>2</sub>*
  integrator.integrator.u | *u(t<sub>1</sub> + dt)*
  integrator.integrator.uprev | *u(t<sub>1</sub>)*
  integrator.integrator.t | *t<sub>1</sub> + dt*
  integrator.integrator.tprev | *t<sub>1</sub>*
  integrator.integrator.k | *k<sub>[t<sub>1</sub>, t<sub>1</sub> + dt]</sub>(u(t<sub>1</sub>), u(t<sub>1</sub> + dt))*

  We then continue at the top in `DelayDiffEq.perform_step!` to calculate *u(t<sub>1</sub> + dt<sub>2</sub>)*.

When we reached the end point, the solution is cleaned by

  * `DelayDiffEq.postamble!`: We update time point `integrator.integrator.t`, value `integrator.integrator.u`, and interpolation data `integrator.integrator.k` with their, maybe more up-to-date, corresponding fields of `integrator`, i.e. with `integrator.t`, `integrator.u`, and `integrator.k`. This method is executed after `OrdinaryDiffEq.loopfooter!`, and hence in the example above we would update `integrator.integrator` with the values *t<sub>1</sub> + dt*, *u(t<sub>1</sub> + dt)*, and *k<sub>[t<sub>1</sub>, t<sub>1</sub> + dt]</sub>(u(t<sub>1</sub>), u(t<sub>1</sub> + dt))*. If the values at time point `integrator.integrator.t` are not contained in the solution of `integrator.integrator` yet, they are added to the solution by `OrdinaryDiffEq.postamble!`.

## Considerations

* Time points `integrator.t`, `integrator.tprev`, `integrator.integrator.t`, and `integrator.integrator.tprev` as well as time step `integrator.dt` do not use pointers, hence they can always be updated by a simple assignment without any special undesired effects.

* FSAL algorithms for in-place problems sometimes use interpolation data `integrator.k` whose first element `integrator.k[1]` points to `integrator.fsalfirst`, which in turn points to a field of `integrator.cache` (see e.g. `BS3Cache`). Since both `integrator` and `integrator.integrator` are initialized with the same cache `integrator.cache` then both `integrator.fsalfirst` and `integrator.integrator.fsalfirst` point to `integrator.cache.fsalfirst`, and both `integrator.k[1]` and `integrator.integrator.k[1]` point to `integrator.cache.fsalfirst`. Hence all these variables act on the same array, as long as they are not forcefully separated. This is a problem since most of the time we would like to have `integrator.k != integrator.integrator.k`, as shown above. Moreover, we want to update the interpolation data `integrator.integrator.k` only after a new estimate *u<sup>n</sup>(t<sub>1</sub> + dt)* whereas all in-place updates to `integrator.k` immediately influence `integrator.integrator.k`. Another issue is that FSAL algorithms for in-place function already update `integrator.fsalfirst` to `integrator.fsallast` in `OrdinaryDiffEq.apply_step!`, and hence if `integrator.k[1]` points to `integrator.fsalfirst` it holds `integrator.k[1] == integrator.fsalfirst == integrator.fsallast` at this point, and even worse, by the above argumentation also `integrator.integrator.k[1] == integrator.fsalfirst == integrator.fsallast` which corrupts the interpolation of `integrator.integrator`.

* As seen, most of the time we should have `integrator.u != integrator.integrator.u` and `integrator.uprev != integrator.integrator.uprev`. If these variable are scalars we could still work with simple assignments, however, we quickly run into trouble if they point to the same arrays. So for arrays the best/easiest/most convenient option seems to be to separate those variables all the time, i.e. to let them point to different arrays all the time, and only update these arrays with `recursivecopy!` when needed.

* At the end of the fixed point iteration we have to reset fields of `integrator.integrator` to the appropriate values in the time interval *[t<sub>0</sub>, t<sub>1</sub>]*. We can easily update the time points by `integrator.integrator.t = integrator.t` and `integrator.integrator.tprev = integrator.tprev`, and reset the value *u(t<sub>1</sub>)* with `integrator.integrator.u = integrator.uprev`. However, *u(t<sub>0</sub>)* and *k<sub>[t<sub>0</sub>, t<sub>1</sub>]</sub>(u(t<sub>0</sub>), u(t<sub>1</sub>))* can not be recovered from any variable, hence we have to cache them. To save allocations it might be desirable to add two additional fields, e.g. `integrator.uprev_cache` and `integrator.k_cache`, to the `DDEIntegrator` such that one can save and reset those values with `recursivecopy!` (if they are arrays, otherwise simple assignments work).

* There is no need in calculating an error estimate of the fixed point iteration during its first iteration. The calculated value just classifies how far *u(t<sub>1</sub> + dt)* differs from *u(t<sub>1</sub>)* or from a value *u(t<sub>1</sub> + dT)* with *dT > dt* that was not accepted in last iteration. Because of this it might actually falsely accept values *u(t<sub>1</sub> + dt)* close to those old values.

* Does the error estimate `integrator.EEst` of the ODE solvers alone prevent that steps with poor error estimate in the fixed point iteration are accepted? The fixed point iteration might just return a value *u(t<sub>1</sub> + dt)* with poor error estimate since it reached the maximum number of iterations. If this is a problem: since both the error estimate of the fixed point iteration and the error estimate `integrator.EEst` of the last iteration of the ODE solver are acceptable if they are not greater than 1, an easy solution would be to set `integrator.EEst` to `max(integrator.EEst, fixedpointEEst)`.

devmotion:
I tried to implement exactly what I wrote down above. It seems to work quite well, I could improve some of the error bounds in the tests significantly. Partly this might be due to the improved extrapolation and due to the fact that the error estimate now combines both the error estimate of the integrator and the error estimate of the fixed-point iteration, which might lead to a higher rate of unaccepted time steps and thus sometimes smaller time steps.

It is even possible to use methods with lazy interpolants (https://github.com/JuliaDiffEq/DelayDiffEq.jl/issues/16) since additional steps are added to the interpolation before values are saved or a new step is calculated. I tested the problem `prob_dde_1delay` with `Vern9`, and the calculated steps are reasonable, however, the interpolation seems to be a bit odd - so there might still be a problem with those methods.

Two other problems I faced:
* It can happen (e.g. in https://github.com/JuliaDiffEq/DelayDiffEq.jl/blob/master/test/unconstrained.jl#L67) that the fixed-point iteration blows up and finally ends up with `u` and `EEst` being `NaN`. `OrdinaryDiffEq.jl` then will try a new step size of value `NaN` in the next execution of `perform_step!`, which will just result in another `u` and `EEst` that are not numbers and hence in an infinite loop of iterations. Thus this has to be fixed and stopped immediately, and therefore the error estimate of the last iteration (which combines both the error estimate of the fixed-point iteration and of the integrator) and the initial interpolation data are cached, the fixed-point iteration is stopped if any `NaN` occur, and both the error estimate (which is responsible for the calculation of the next time step) and the interpolation data (which is used to calculate the next step) are reset to the cached values. With this change all tests can be run successfully.
* I also tested for other methods like `Tsit5`, `DP8`, and among others the Verner methods whether they produce the same results for in-place and not in-place functions. Out of the methods I tested only `Vern7` leads to different time steps and slightly different error estimates. Thus I tested whether this is reproducible with an ODE system, and indeed `Vern7` produces different results for in-place and not in-place functions also with ODEs. So probably there is an error in one of the implementations of `Vern7`.

Moreover, with this changed implementation I get the best error estimates for my toy example so far; before (https://github.com/JuliaDiffEq/DelayDiffEq.jl/issues/15#issue-241997704) the error estimates were

```julia
julia> sol.errors
Dict{Symbol,Float64} with 3 entries:
  :l∞    => 3.33289e-5
  :final => 1.86351e-5
  :l2    => 1.40652e-5

julia> sol2.errors
Dict{Symbol,Float64} with 3 entries:
  :l∞    => 3.31446e-5
  :final => 1.86005e-5
  :l2    => 1.401e-5
```

now they are

```julia
julia> sol.errors
Dict{Symbol,Float64} with 3 entries:
  :l∞    => 2.98268e-5
  :final => 1.7883e-5
  :l2    => 1.30246e-5

julia> sol2.errors
Dict{Symbol,Float64} with 3 entries:
  :l∞    => 2.98268e-5
  :final => 1.7883e-5
  :l2    => 1.30246e-5
```

The tests pass locally but need https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/80 which is merged but not released yet.

devmotion:
Tests pass on Windows 64bit, Linux, and Mac - but the stricter error bounds lead to failures on Windows 32bit. Is there a workaround without sacrificing the better bounds completely?

devmotion:
I still think that the outline above summarises what we want to do - but does the extrapolation really work correctly?

I found some papers regarding our approach, e.g. https://link.springer.com/article/10.1007/BF02072017 and http://www.sciencedirect.com/science/article/pii/S0168927497000226, and because of the obvious formula for `Θ` on p. 220 of Bellen I'm a bit worried that we might extrapolate incorrectly. My main concern is the top part about `perform_step!`, and especially the values we use to calculate the extrapolation.

First of all, every extra- and intrapolation depends on both the last time point of the `integrator.integrator.sol` and on the current and previous values of `integrator.integrator` (since `current_interpolant` calculates `Θ` as `(t-integrator.integrator.tprev)/integrator.dt`. Everything should be fine for any `t` before the current end point of the solution that just has to be interpolated, i.e. for all `t` in *[t<sub>0</sub>, t<sub>1</sub>]* in the example above. For these time points the history function will just use the interpolation of the solution `integrator.integrator.sol`, so we should not have to worry about it. But for any `t` after the current end point of the solution, i.e. any `t` that is greater than *t<sub>1</sub>* in the example above, the history function will just call `current_interpolant`, which in turn defines `Θ` as `(t-integrator.integrator.tprev)/integrator.integrator.dt`.

So my question is: don't we have to update `integrator.integrator.tprev` and `integrator.integrator.dt` already to *t<sub>1</sub>* (the last time point of the solution) and *dt* (the next time step) to calculate a meaningful `Θ`? As far as I see, this also would not affect any interpolations since they only depend on the solution `integrator.integrator.sol` and it would correspond to the formula given by Bellen.

But maybe I'm also just confused right now... It's really tricky to get it right it seems.

ChrisRackauckas:
>First of all, every extra- and intrapolation depends on both the last time point of the integrator.integrator.sol and on the current and previous values of integrator.integrator (since current_interpolant calculates Θ as (t-integrator.integrator.tprev)/integrator.dt. Everything should be fine for any t before the current end point of the solution that just has to be interpolated, i.e. for all t in [t0, t1] in the example above. For these time points the history function will just use the interpolation of the solution integrator.integrator.sol, so we should not have to worry about it. But for any t after the current end point of the solution, i.e. any t that is greater than t1 in the example above, the history function will just call current_interpolant, which in turn defines Θ as (t-integrator.integrator.tprev)/integrator.integrator.dt.

We just get a `Θ > 1` which is perfectly fine. That form of extrapolation is common for producing the starting values of an implicit RK method. Then the interval is shifted using these starting values to generate a continuous solution over the next interval, and interpolating in there then makes `Θ <= 1` for every interpolation.

devmotion:
Ah OK, good to know. I am just worried there might still be some failures in the algorithm - in particular since suddenly the iterations were exploding for some special cases. And I was thinking whether one can stop the fixed point iterations as soon as the error estimate increases between two subsequent iterations.

ChrisRackauckas:
>And I was thinking whether one can stop the fixed point iterations as soon as the error estimate increases between two subsequent iterations.

We should switch over to Anderson first, and this could probably be something in NLsolve.jl and we then interpret a retcode.

ChrisRackauckas:
Where exactly are we on this PR? Ready for review again?

devmotion:
Finally I fixed the conflicts, so it is ready for review again. Took me awhile since I ran into https://github.com/ajkeller34/Unitful.jl/issues/95 and I figured out that in fact the interpolation was done incorrectly - `integrator.integrator.dt` was never updated. This probably also caused the exploding fixed-point iterations, since they are gone now; but I did not remove the corresponding code, so the implementation should still be able to deal with exploding fixed-point iterations (in case they might happen).

Moreover, due to this change lazy interpolants do not work anymore, so for now I completely removed the code which was supposed to add additional steps and handle these algorithms.

devmotion:
Again HDF5 build errors on Windows 32bit but it seems these issues are common and known: https://github.com/JuliaPackaging/WinRPM.jl/issues/120, https://github.com/JuliaPackaging/WinRPM.jl/issues/118, https://github.com/JuliaPackaging/WinRPM.jl/issues/107, and https://github.com/JuliaPackaging/WinRPM.jl/pull/117

src/integrator_utils.jl
```diff
@@ -229,3 +229,28 @@ function build_solution_interpolation(integrator::DDEIntegrator, sol::DiffEqArra
...
+    update_ode_integrator!(integrator::DDEIntegrator)
+
+Update ODE integrator of `integrator` to current time interval, values and interpolation
+data of `integrator`.
+"""
+function update_ode_integrator!(integrator::DDEIntegrator)
+    # update time interval of ODE integrator
+    integrator.integrator.t = integrator.t
+    integrator.integrator.tprev = integrator.tprev
+    integrator.integrator.dt = integrator.integrator.t - integrator.integrator.tprev
```
ChrisRackauckas: why not `integrator.dt`? This could be open to floating point issues.

src/solve.jl
```diff
@@ -75,7 +75,7 @@ function init(prob::AbstractDDEProblem{uType,tType,lType,isinplace}, alg::algTyp
                                                        ode_prob,
                                                        OrdinaryDiffEq.alg_order(alg)))
     end
-    integrator.dt = dt
+    integrator.dt = zero(dt)
```
ChrisRackauckas: won't this override the init dt choice from the user?

devmotion: No, this just assures that the ODE integrator always satisfies `tprev + dt == t`. The DDE integrator is still initialized with the init dt choice from the user (or the calculated init dt).

ChrisRackauckas: Setting `dt = t - tprev` does not assure that the ODE integrator satisfies `tprev + dt == t`

devmotion: I'm not sure I understand you. Of course, for all other steps (besides the initial step) lines https://github.com/devmotion/DelayDiffEq.jl/blob/8008e8f2d0837c944e7a7093a948a4fb094f0bb5/src/integrator_utils.jl#L239-L242 and https://github.com/devmotion/DelayDiffEq.jl/blob/8008e8f2d0837c944e7a7093a948a4fb094f0bb5/src/integrator_interface.jl#L70-L72 guarantee that `t`, `tprev`, and `dt`of the ODE integrator `integrator.integrator` belong to the same time interval which is important for correct interpolation. So setting `dt` to zero initially (only for `integrator.integrator`, the DDE integrator still uses the supplied or calculated `dt`!) since `tprev = t` initially is not needed for any calculations but ensures this consistent relationship of these three fields throughout the whole algorithm.

Of course, because of floating point issues the subtraction is a bad idea, and hence I corrected this in the last commit.

ChrisRackauckas: Does this actually have a practical impact, or just makes things nicer because then `tprev + dt == t`?

devmotion: This specific line here is without any practical impact, I would say. For a new initialised DDE integrator `dde_int` it leads to `isnan.(dde_int.f(t, u))` for any `t > minimum(prob.lags)` since any interpolation `current_interpolant` is also not a number - but it does not influence the calculations since the initial step is always at most of size of the minimal delay (assured since the discontinuities are tstops).

However, for all other steps it makes a huge difference (by the addition of these updates I got rid of the exploding fixed-point iterations in some tests!), since any interpolation that is used in the function evaluation and hence also the calculation of the next steps is based on the current time interval of the ODE integrator, and in particular also on the current value of `dt` (see https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/dense/generic_dense.jl#L40 and similar lines). These lines also show that for an initial dt of zero `Θ` is calculated by a division by zero, and hence this explains why the interpolation does not work initially.

So this specific line could be removed but I thought it would be preferable to keep `dt`, `t`, and `tprev` synchronized all the time.

ChrisRackauckas: >However, for all other steps it makes a huge difference

Not this line though, just the PR in general?

>So this specific line could be removed but I thought it would be preferable to keep dt, t, and tprev synchronized all the time.

Nah, it makes sense. Just was curious about the practical effect. In theory it would do a constant extrapolation if `dt=0`, I just was wondering if that ever came into play.

devmotion: > Not this line though, just the PR in general?

Yes just the PR in general.

> Nah, it makes sense. Just was curious about the practical effect. In theory it would do a constant extrapolation if dt=0, I just was wondering if that ever came into play.

But it is not implemented yet, is it? As far as I see, all calls of `current_interpolant` and `current_interpolant!` (and also of `current_extrapolant` and `current_extrapolant!`) will calculate a `Θ` that is not a number for `dt=0` - which will always lead to interpolations/extrapolations that are not a number, I guess.

ChrisRackauckas: >But it is not implemented yet, is it? As far as I see, all calls of current_interpolant and current_interpolant! (and also of current_extrapolant and current_extrapolant!) will calculate a Θ that is not a number for dt=0 - which will always lead to interpolations/extrapolations that are not a number, I guess.

But almost all of the interpolations have `u0 + dt* ...`, so it'll "accidentally" work out as a constant extrapolation. I think just the SSPRK methods don't. But I don't think we're actually hitting this case so it probably doesn't come up. Probably worthwhile to address in another PR if it does, with the SSPRK methods/interpolants as the test case which doesn't have the accidental workaround.

devmotion: > But almost all of the interpolations have u0 + dt* ..., so it'll "accidentally" work out as a constant extrapolation.

I guess it won't, and it didn't in my tests with BS3 and Tsit5. The problem is that the expression in brackets after `u0 + dt*` is not a number in these cases; an example is line https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/dense/interpolants.jl#L123.

ChrisRackauckas: oh, 0*NaN = NaN? I never knew that. Okay, then I guess that doesn't happen.

PR #18, opened 2017-07-14
Add support for saveat and save_idxs to MethodOfSteps algorithm

devmotion:
Hi!

Currently the keyword `save_idxs` does usually not work with the MethodOfSteps algorithm. Often additional components are necessary during evaluation of the history function, and even the indices of included components do not match the indices in the original delay differential equations.

Thus I added an additional solution object to the DDE integrator such that it is possible to output only a selection of components. Internally for evaluations of the history function always a full solution is used. Of course, it is possible to extract the desired components after all iterations but it seemed more natural to me to take advantage of the implementations in OrdinaryDiffEq.jl (it is necessary to loosen the type restriction in `savevalues!` there!). With an iterator interface similar to OrdinaryDiffEq.jl the output solution would contain only the desired components in every step automatically.

I already included support for `saveat` as well: because of the usually sparse solutions with this option and hence increased number of fixed-point iterations the use of `saveat` can increase computation time dramatically at the moment. However, with this change computation time should not increase.

So, maybe this suggestion is helpful. I added some tests that lead to errors on the master branch.

### DIFF
--- a/src/integrator_interface.jl
+++ b/src/integrator_interface.jl
@@ -9,6 +9,9 @@ function savevalues!(integrator::DDEIntegrator, force_save=false)
     integrator.integrator.k = integrator.k
     integrator.integrator.t = integrator.t
 
+    # update solution of DDE integrator
+    invoke(savevalues!, Tuple{AbstractDDEIntegrator,Bool}, integrator, force_save)
+
     # add steps for interpolation to ODE integrator when needed
     OrdinaryDiffEq.ode_addsteps!(integrator.integrator, integrator.f)
 
@@ -26,6 +29,9 @@ function postamble!(integrator::DDEIntegrator)
     integrator.integrator.u = integrator.u
     integrator.integrator.k = integrator.k
     integrator.integrator.t = integrator.t
+
+    # clean up solution of DDE integrator
+    OrdinaryDiffEq.postamble!(integrator)
 
     # clean up solution of ODE integrator
     OrdinaryDiffEq.postamble!(integrator.integrator)
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -1,8 +1,9 @@
 function init(prob::AbstractDDEProblem{uType,tType,lType,isinplace}, alg::algType,
               timeseries_init=uType[], ts_init=tType[], ks_init=[];
               d_discontinuities=tType[], dtmax=tType(7*minimum(prob.lags)), dt=zero(tType),
-              kwargs...) where {uType,tType,isinplace,
-                                algType<:AbstractMethodOfStepsAlgorithm,lType}
+              saveat=tType[], tstops=tType[], save_idxs=nothing,
+              save_everystep=isempty(saveat), save_start=true, kwargs...) where
+    {uType,tType,isinplace,algType<:AbstractMethodOfStepsAlgorithm,lType}
 
     # add lag locations to discontinuities vector
     d_discontinuities = [d_discontinuities; compute_discontinuity_tree(prob.lags, alg,
@@ -16,6 +17,8 @@ function init(prob::AbstractDDEProblem{uType,tType,lType,isinplace}, alg::algTyp
 
     # bootstrap the integrator using an ODE problem, but do not initialize it since
     # ODE solvers only accept functions f(t,u,du) or f(t,u) without history function
+    # save every step and every index in the solution of the integrator, since they are
+    # required for evaluation of past values during integration
     ode_prob = ODEProblem(prob.f, prob.u0, prob.tspan; iip=isinplace)
     integrator = init(ode_prob, alg.alg; dt=1, initialize_integrator=false,
                       d_discontinuities=d_discontinuities, dtmax=dtmax, kwargs...)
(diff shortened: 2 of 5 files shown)


src/solve.jl
```diff
@@ -1,8 +1,9 @@
 function init(prob::AbstractDDEProblem{uType,tType,lType,isinplace}, alg::algType,
               timeseries_init=uType[], ts_init=tType[], ks_init=[];
               d_discontinuities=tType[], dtmax=tType(7*minimum(prob.lags)), dt=zero(tType),
-              kwargs...) where {uType,tType,isinplace,
-                                algType<:AbstractMethodOfStepsAlgorithm,lType}
+              saveat=tType[], tstops=tType[], save_idxs=nothing,
```
ChrisRackauckas: why can't these be handled by the splatted kwards?

src/integrator_interface.jl
```diff
@@ -9,6 +9,9 @@ function savevalues!(integrator::DDEIntegrator, force_save=false)
     integrator.integrator.k = integrator.k
     integrator.integrator.t = integrator.t
 
+    # update solution of DDE integrator
+    invoke(savevalues!, Tuple{AbstractDDEIntegrator,Bool}, integrator, force_save)
```
ChrisRackauckas: isn't this pointing to the same solution as the ODEIntegrator? Why does it need its own save?

src/solve.jl
```diff
@@ -32,60 +35,92 @@ function init(prob::AbstractDDEProblem{uType,tType,lType,isinplace}, alg::algTyp
     end
 
     if typeof(alg.alg) <: OrdinaryDiffEqCompositeAlgorithm
-        interp_data = OrdinaryDiffEq.CompositeInterpolationData(integrator.sol.interp,
-                                                                interp_f)
+        ode_interp_data = OrdinaryDiffEq.CompositeInterpolationData(integrator.sol.interp,
```
ChrisRackauckas: this is making it hard to review. Is half of what's going on an unrelated renaming?

src/solve.jl
devmotion: Since the internal solution of the ODE integrator is supposed to be always a full solution with all components that saves every step. The solution of the DDE integrator can be modified by these keywords.

ChrisRackauckas: But then do we ever gain anything by having double the solution? It's not like that makes it more memory efficient if the ODE solution is already saving everything.

src/integrator_interface.jl
devmotion: No, these are two different solutions. One reason is that often other components of the history function have to be evaluated as well, e.g. if the system would be ``x'(t) = y(t-1), y'(t) = x(t-2) + y(t)`` then one can not evaluate ``x(t-2)`` if one just restricts the solution of the ODE integrator to ``y`` (as currently done).

src/solve.jl
devmotion: No, just ``ode_interp_data`` and ``ode_sol`` are renamed since ``dde_interp_data`` and ``dde_sol`` are added.

src/solve.jl
devmotion: As mentioned above, one could of course just filter e.g. components or time steps afterwards. But I thought it might be better to mostly use the already existing methods in OrdinaryDiffEq.jl, instead of implementing these things in a completely different way for DDEs. It's just that these keywords are not usable at the moment, and since I would like to retrieve only specific components at specific time points  I came up with this first idea.

ChrisRackauckas: The thing that worries me here though is it seems this then requires doubling everything in a way that it didn't before. Doubling the solution type and doubling the saving will have a big impact on the runtime.

devmotion: Yes that's clearly a disadvantage of this approach. Since it only concerns the saving and not the calculations or anything else, I thought it could still be an option. In constrast to ODEs it seems that, at least in general, the ODE integrator really should always save all time steps and all components.

devmotion: So I think the keyword filtering is usually necessary, but maybe it's smarter to build an output solution with the desired components and time points only at the end.

ChrisRackauckas: Can this be done more cheaply?

ChrisRackauckas:
I think this PR is a lesser priority than the others. I'll take another look through the others and start merging those ASAP, but this will take more thought and benchmarks.

src/solve.jl
devmotion: Building an output solution at the end? I'm not sure, I will try to compare it.
Another idea would be to add an option `calc_idxs` that already restricts all calculations to a subset of components and/or intermediate purging of the solution, since one probably only needs interpolation in the interval `[t-maximum(prob.lags), t]`.

devmotion:
Yes, I also think this needs more discussion and benchmarks. I will try to test some of the ideas in https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/18#discussion_r127579809 in the meantime.

devmotion:
I set up a first implementation that creates an output solution, suitable to `save_idxs`, `saveat`, `save_everystep`, and `save_start`, after all steps and calculations are done. It is available at https://github.com/devmotion/DelayDiffEq.jl/tree/final_solution. In its current state the ODE integrator always saves every step and every component. The output solution still contains the dense interpolation, but to reduce it one could/should add an option which disables that behaviour.

devmotion:
I extended the other implementation in branch `final_solution` (https://github.com/devmotion/DelayDiffEq.jl/tree/final_solution) such that it is also possible to discard the dense interpolation in the output solution.

Moreover, I ran some small benchmarks and compared `master`, `saveat_save_idxs` (this PR), and `final_solution`. The benchmarks together with the complete results of my local run are available at https://gist.github.com/devmotion/97a1c30df6fe7e0774b6456cc0240a24. Just a short summary here:
1. First I compared the usage of keywords `save_idxs = nothing` and `save_idxs = 1:3` on `master`, `saveat_save_idxs`, and `final_solution`:
   * `master` branch has least memory estimates
   * `final_solution` almost equal to `master` regarding memory estimates and time, except for memory estimates for `save_idxs = 1:3`
   * memory estimates on `master` branch are always reduced for `save_idxs = 1:3` (seems to save memory in ODE integrator - but why should this be the case?)
   * `final_solution` always better than `saveat_save_idxs` with respect to memory estimates and time
2. Next I compared `saveat_save_idxs` and `final_solution` with `save_idxs = [1]`, `save_idxs = [3]`, `save_idxs = 1:2`, and `save_idxs = 2:3` (does not run on `master`):
   * As expected, benchmarks with `save_idxs = [1]` and `save_idxs = [3]` are very similar, as well as benchmarks with `save_idxs = 1:2` and `save_idxs = 2:3`
   * Times of branches `saveat_save_idxs` and `final_solution` very similar
   * Memory estimates of `final_solution` slightly less than memory estimates of `saveat_save_idxs`

So I would suggest using the approach of `final_solution`, although the implementation seems slightly more complicated. `final_solution` also introduces (almost) no regression and overhead for default parameters.

devmotion:
Closed in favor of https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/19.

PR #19, opened 2017-07-24
Enable use of keywords save_idxs, saveat, save_start, save_everystep, and dense

devmotion:
This PR tries to solve the same issues as https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/18. Parameter estimation would be simplified when these keywords can be applied. Currently, the use of `save_idxs` and `save_start` is (usually) not possible, and the use of `saveat`, `save_everystep`, and `dense` leads to incorrect solutions. In contrast to the old PR this implementation does not use a second, separate solution with specified indices and time points immediately, but only after all calculations are done.

Moreover, if the final solution is not dense and does only contain time points specified by `saveat`, the ODE solution is already reduced during the calculations, i.e. time points that are not required for accurate interpolation of the time points specified by `saveat` are removed as soon as they are not needed anymore to correctly calculate the next step.

Some simple benchmarks comparing current master, the old PR, and this PR are available at https://gist.github.com/devmotion/97a1c30df6fe7e0774b6456cc0240a24.


devmotion:
I do not know why but apparently the build on Windows 32bit again fails because of some error bounds, (see https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/17#issuecomment-317173759):

https://ci.appveyor.com/project/ChrisRackauckas/delaydiffeq-jl/build/1.0.85/job/i2n6fykov52cueo0

I did not even change those bounds... Tests pass on all other architectures.

src/solve.jl
```diff
@@ -171,23 +239,179 @@ function solve!(integrator::DDEIntegrator)
     # clean up solution
     postamble!(integrator)
 
-    # calculate and add errors to solution if analytic solution to problem exists
-    # can not use same mechanism as for ODE problems since analytic solutions for DDE depend
-    # on initial value u0
+    if integrator.opts.save_everystep && isempty(integrator.opts.saveat)
```
ChrisRackauckas: Can this be pulled out to a function call on the integrator? It would make `solve` a bit more legible.

src/solve.jl
```diff
@@ -171,23 +239,179 @@ function solve!(integrator::DDEIntegrator)
...
+                          for t in t_sol]
+        end
         errors = Dict{Symbol,eltype(integrator.u)}()
-        sol = build_solution(integrator.sol::AbstractODESolution, u_analytic, errors)
+    else
+        u_analytic = nothing
+        errors = nothing
+    end
+
+    if integrator.opts.dense
```
ChrisRackauckas: Pull this out to a function too?

src/utils.jl
```diff
@@ -29,3 +29,69 @@ function compute_discontinuity_tree(lags, alg, start_val)
...
+            # do not remove time point if it is lower bound of a required interpolation
+            # interval
+            needed = needed || (!(isempty(integrator.saveat)) &&
+                                integrator.saveiter + 1 < integrator.integrator.saveiter &&
+                                t_sol < top(integrator.saveat) <
+                                integrator.integrator.sol.t[integrator.saveiter + 2])
+
+            if !needed
+                # delete not required time point, function value, and interpolation data
+                deleteat!(integrator.integrator.sol.t, integrator.saveiter + 1)
```
ChrisRackauckas: won't deleateat! be slow to do one-by-one, and it would be much faster to just do the change at the end? I am not sure why this is needed if you're also doing the other thing at the end.

devmotion: It's not necessary to delete these entries, but I wanted to check if there is any benefit in deleting unneeded entries already during the algorithm.
The problem is that the entries that can be deleted do not have to be successive entries - some entries might be needed to calculate interpolated values of the output solution whereas others can be removed completely... Maybe one could completely remove these entries and only keep the interpolated values - but I do not know if this could cause other problems....

ChrisRackauckas: I'd run timings on it. I know Julia amortizes `push!`, but I think it might re-allocate with eery `deleteat!` and subsequently ignore the amortization, making it quite expensive to `delateat!` very often. I would only want to use this code path then in cases where I have a really tight RAM limit that I am hitting. It would probably be good to keep it as an option.

ChrisRackauckas:
Cool, that was it. Tests pass. So after moving those pieces out to a function and making an option for the `deleteat!` we can merge.

devmotion:
Currently HDF5 has build errors on Windows 32bit:
```
INFO: Downloading: hdf5
WARNING: Unknown download failure, error code: 2148270086
WARNING: Retry 1/5 downloading: https://cache.julialang.org/http://download.opensuse.org/repositories/windows:/mingw:/win32/openSUSE_Leap_42.2/noarch/mingw32-hdf5-1.8.13-4.13.noarch.rpm
WARNING: Unknown download failure, error code: 2148270086
WARNING: Retry 2/5 downloading: https://cache.julialang.org/http://download.opensuse.org/repositories/windows:/mingw:/win32/openSUSE_Leap_42.2/noarch/mingw32-hdf5-1.8.13-4.13.noarch.rpm
WARNING: Unknown download failure, error code: 2148270086
WARNING: Retry 3/5 downloading: https://cache.julialang.org/http://download.opensuse.org/repositories/windows:/mingw:/win32/openSUSE_Leap_42.2/noarch/mingw32-hdf5-1.8.13-4.13.noarch.rpm
WARNING: Unknown download failure, error code: 2148270086
WARNING: Retry 4/5 downloading: https://cache.julialang.org/http://download.opensuse.org/repositories/windows:/mingw:/win32/openSUSE_Leap_42.2/noarch/mingw32-hdf5-1.8.13-4.13.noarch.rpm
WARNING: Unknown download failure, error code: 2148270086
WARNING: Retry 5/5 downloading: https://cache.julialang.org/http://download.opensuse.org/repositories/windows:/mingw:/win32/openSUSE_Leap_42.2/noarch/mingw32-hdf5-1.8.13-4.13.noarch.rpm
INFO: try running WinRPM.update() and retrying the install
================================[ ERROR: HDF5 ]=================================
LoadError: failed to download hdf5 0 from https://cache.julialang.org/http://download.opensuse.org/repositories/windows:/mingw:/win32/openSUSE_Leap_42.2/noarch/mingw32-hdf5-1.8.13-4.13.noarch.rpm.
while loading C:\Users\appveyor\.julia\v0.6\HDF5\deps\build.jl, in expression starting on line 39
```
I noticed the same errors some days ago when I tried to debug the Windows 32bit error but it was gone after some hours. So probably we can just try to restart the tests til it works...

devmotion:
At least Travis CI builds pass but there's still the download error on Windows 32bit...

ChrisRackauckas:
well since it did run Win32 once and pass I'm pretty confident it's fixed. Since all of the changes are made I'll merge now. Of course, this will be re-run in the future, so if the test fails when the caching gets fixed we can open it up again.

PR #23, opened 2017-08-01
Fix wrongly defined lags in unconstrained.jl

devmotion:
In `unconstrained.jl` `lags` was not defined, so the tests wrongly used `lags = [1//5, 1//2]` from `discont_tree_test.jl`. This did not cause any trouble because there are no error bound for the affected example.

Tests will fail because both MuladdMacro and DiffEqBase export `@muladd`. I would suggest tagging a new version of DiffEqBase, which then can be added as lower bound to DelayDiffEq (and also OrdinaryDiffEq, and StochasticDiffEq, I guess).

### DIFF
--- a/test/discont_tree_test.jl
+++ b/test/discont_tree_test.jl
@@ -1,4 +1,4 @@
-using DelayDiffEq, OrdinaryDiffEq, Base.Test
+using DelayDiffEq, OrdinaryDiffEq, DataStructures, Base.Test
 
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
Yeah, maybe the interface should somehow be changed in a way that requires that the lags line up, since right now it's on a "trust" system instead of a guarantee. I'm not sure how to do it well without vastly overcomplicating things though.

PR #28, opened 2017-08-13
Update integrator type to OrdinaryDiffEq v2.15.0

devmotion:
This updates DelayDiffEq to the changed integrator type and its options in OrdinaryDiffEq v2.15.0. Tests pass locally but have to be restarted if https://github.com/JuliaLang/METADATA.jl/pull/10704 is merged.


ChrisRackauckas:
Yes, this needs to be tagged and we need version bounds in METADATA. Thanks.

PR #30, opened 2017-08-24
Lazy interpolants

devmotion:
These changes enable the use of methods with lazy interpolants such as Verner methods. Fixes https://github.com/JuliaDiffEq/DelayDiffEq.jl/issues/16.

### DIFF
--- a/src/integrator_interface.jl
+++ b/src/integrator_interface.jl
@@ -55,6 +55,9 @@ Calculate next step of `integrator`.
     recursivecopy!(integrator.k_cache, integrator.k)
     integrator.integrator.EEst = integrator.EEst
 
+    # add additional interpolation steps
+    OrdinaryDiffEq.ode_addsteps!(integrator.integrator, integrator.f)
+
     # perform always at least one calculation
     perform_step!(integrator, integrator.cache)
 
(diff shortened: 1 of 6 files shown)


ChrisRackauckas:
I just read the title and got excited.

src/integrator_interface.jl
```diff
@@ -81,7 +84,8 @@ Calculate next step of `integrator`.
         else
             integrator.uprev_cache = integrator.integrator.uprev
         end
-        recursivecopy!(integrator.k_integrator_cache, integrator.integrator.k)
+        recursivecopy!(integrator.k_integrator_cache,
+                       view(integrator.integrator.k, 1:integrator.kshortsize))
```
ChrisRackauckas: spacing

src/integrator_interface.jl
```diff
@@ -161,9 +169,16 @@ Calculate next step of `integrator`.
             integrator.integrator.u = integrator.uprev
             integrator.integrator.uprev = integrator.uprev_cache
         end
-        recursivecopy!(integrator.integrator.k, integrator.k_integrator_cache)
+        recursivecopy!(view(integrator.integrator.k, 1:integrator.kshortsize),
+                       integrator.k_integrator_cache)
+
+        # remove additional interpolation steps
+        resize!(integrator.k, integrator.kshortsize)
```
ChrisRackauckas: I think you might as well keep them if you've already calculated them. The reason why they're lazy is because usually it's a lot more work to calculate those steps and not a memory thing.

ChrisRackauckas:
Looks good other than the removal.

src/integrator_interface.jl
devmotion: But even without resizing `integrator.k` is different from the `integrator.integrator.k` that is calculated later in https://github.com/devmotion/DelayDiffEq.jl/blob/4c1e4d1773d565e002e18ac78147034f6d538384/src/integrator_utils.jl#L258. The additional interpolation steps in `integrator.k` are still the interpolation steps of the previous fixed point iteration, so when `resize!` is removed here we would then need an additional `ode_addsteps!`.

ChrisRackauckas: I see. Makes sense.

PR #31, opened 2017-08-25
Enable SDIRK integrators

devmotion:
This enables SDIRK integrators. Currently they do not work since the derivative wrappers, created during creation of the ODE integrator, wrap function `prob.f` which requires 4 arguments. Hence I created an additional cache which wraps function `dde_f` of the DDE integrator with 3 arguments.

By the way, Euler methods are very slow which is caused by the incorrect implementation of divided differences IMO. Changing line
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/integrators/sdirk_integrators.jl#L230
to
```julia
@. tmp = r*abs((u - uprev)/dt1 - (uprev - uprev2)/dt2)
```
(which corresponds to the definition of divided differences, I guess) reduces both timings and allocations to similar values as in other SDIRK methods.

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -124,7 +124,7 @@ function init(prob::AbstractDDEProblem{uType,tType,lType,isinplace}, alg::algTyp
         fixedpoint_reltol_internal = map(uEltypeNoUnits, alg.fixedpoint_reltol)
     end
 
-    # create separate copies u and uprev, not pointing integrator.u or integrator.uprev,
+    # create separate copies u and uprev, not pointing to integrator.u or integrator.uprev,
     # containers for residuals and to cache uprev with correct dimensions and types
     # in particular for calculations with units residuals have to be unitless
     if typeof(integrator.u) <: AbstractArray
@@ -137,6 +137,17 @@ function init(prob::AbstractDDEProblem{uType,tType,lType,isinplace}, alg::algTyp
         uprev = deepcopy(integrator.uprev)
         resid = one(uEltypeNoUnits)
         uprev_cache = oneunit(eltype(uType))
+    end
+
+    # create uprev2 in same way as in OrdinaryDiffEq
+    if integrator.uprev === integrator.uprev2
+        uprev2 = uprev
+    else
+        if uType <: Array
+            uprev2 = copy(uprev)
+        else
+            uprev2 = deepcopy(uprev)
+        end
     end
 
     # define caches for interpolation data
@@ -169,45 +180,15 @@ function init(prob::AbstractDDEProblem{uType,tType,lType,isinplace}, alg::algTyp
         save_idxs = nothing # prevents indexing of ODE solution and saves memory
     end
 
-    # new cache required for methods that wrap derivatives of f such as Rosenbrock
-    # create uprev2 in same way as in OrdinaryDiffEq
-    if integrator.uprev === integrator.uprev2
-        uprev2 = uprev
-    else
-        if uType <: Array
-            uprev2 = copy(uprev)
-        else
-            uprev2 = deepcopy(uprev)
-        end
-    end
-    dde_cache = OrdinaryDiffEq.alg_cache(alg.alg, u, integrator.rate_prototype,
-                                         uEltypeNoUnits, tTypeNoUnits, uprev, uprev2, dde_f,
-                                         integrator.t, dt, integrator.opts.reltol,
-                                         Val{isinplace})
-
-    # create solution of DDE integrator with interpolation that uses new cache
+    # new cache with updated u, uprev, uprev2, and function f
     if typeof(alg.alg) <: OrdinaryDiffEq.OrdinaryDiffEqCompositeAlgorithm
-        dde_interp_data = OrdinaryDiffEq.CompositeInterpolationData(
-            dde_f, interp_data.timeseries, interp_data.ts, interp_data.ks,
-            interp_data.alg_choice, interp_data.notsaveat_idxs, interp_data.dense,
-            dde_cache)
+      caches = map((x,y) -> build_linked_cache(x, y, u, uprev, uprev2, dde_f,
+                                               prob.tspan[1], dt),
+                   integrator.cache.caches, alg.alg.algs)
+      dde_cache = OrdinaryDiffEq.CompositeCache(caches, alg.alg.choice_function, 1)
     else
-        dde_interp_data = OrdinaryDiffEq.InterpolationData(
-            dde_f, interp_data.timeseries, interp_data.ts, interp_data.ks,
-            interp_data.notsaveat_idxs, interp_data.dense, dde_cache)
-    end
-
-    if typeof(alg.alg) <: OrdinaryDiffEq.OrdinaryDiffEqCompositeAlgorithm
-        dde_sol = build_solution(prob, integrator.sol.alg, integrator.sol.t,
-                                 integrator.sol.u, dense=integrator.sol.dense,
-                                 k=integrator.sol.k, interp=dde_interp_data,
-                                 alg_choice=integrator.sol.alg_choice,
-                                 calculate_error = false)
-    else
-        dde_sol = build_solution(prob, integrator.sol.alg, integrator.sol.t,
-                                 integrator.sol.u, dense=integrator.sol.dense,
-                                 k=integrator.sol.k, interp=dde_interp_data,
-                                 calculate_error = false)
+      dde_cache = build_linked_cache(integrator.cache, alg.alg, u, uprev, uprev2, dde_f,
+                                     prob.tspan[1], dt)
     end
 
     # separate options of integrator and ODE integrator since ODE integrator always saves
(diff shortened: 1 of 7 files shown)


src/solve.jl
```diff
@@ -169,6 +169,47 @@ function init(prob::AbstractDDEProblem{uType,tType,lType,isinplace}, alg::algTyp
...
+    if integrator.uprev === integrator.uprev2
+        uprev2 = uprev
+    else
+        if uType <: Array
+            uprev2 = copy(uprev)
+        else
+            uprev2 = deepcopy(uprev)
+        end
+    end
+    dde_cache = OrdinaryDiffEq.alg_cache(alg.alg, u, integrator.rate_prototype,
```
ChrisRackauckas: This is going to double the allocations though. Is there no better way? Why can't the derivative wrapper use the `f` from the delay equation, just differently?

ChrisRackauckas:
Why not just have `f` passed to the caches, and then have the derivative wrapper use that?

ChrisRackauckas:
>@. tmp = r*abs((u - uprev)/dt1 - (uprev - uprev2)/dt2)

Oh, did I do divided additions? Can you make this a PR?

test/sdirk_integrators.jl
```diff
@@ -0,0 +1,25 @@
+using DelayDiffEq, DiffEqProblemLibrary, Base.Test
+
+prob = prob_dde_1delay(1.0)
+
+# ODE algorithms
+algs = [GenericImplicitEuler(), GenericTrapezoid(),
```
ChrisRackauckas: Why not test any Rosenbrocks?

devmotion: There are additional problems with Rosenbrocks, they still do not work.

src/solve.jl
devmotion: At the point when the cache and hence also the derivative wrappers are created the `f` with 3 arguments that implicitly contains the history function is not existent yet. If we just pass `f` (as currently done) then we somehow would also have to provide the history function every time the wrappers are called. Is there an easy way to do that? I didn't know how to solve this problem easily in another way.

devmotion:
I'll prepare a PR regarding divided differences, and I also provide a fix for a typo in the not in-place version of `KenCarp4`.

ChrisRackauckas:
There's gotta be a good way to make it alias the caches via a generated function or something? Doubling the cache would definitely be a huge hit to allocations.

devmotion:
I guess that might be possible. I will see whether I can get rid of these allocations.

devmotion:
I'll have another look at it tomorrow, but so far it seems a bit ugly since one has to make sure that after aliasing still all interdependencies in e.g. lines
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/caches/rosenbrock_caches.jl#L88-L99
are satisfied. In particular it seems difficult to update `jac_config` correctly.

Moreover, I shortly checked `Rosenbrock23` again, and the in-place method fails because of some problem with `ForwardDiff`:
```julia
julia> sol = solve(prob, MethodOfSteps(Rosenbrock23()))
ERROR: MethodError: Cannot `convert` an object of type ForwardDiff.Dual{ForwardDiff.Tag{OrdinaryDiffEq.TimeGradientWrapper{OrdinaryDiffEq.VectorF{DelayDiffEq.##27#32{DiffEqBase.DDEProblem{Array{Float64,1},Float64,Array{Int64,1},Void,true,DiffEqProblemLibrary.##24#25,DiffEqProblemLibrary.##26#27,Void,UniformScaling{Int64}}},Tuple{Int64}},Array{Float64,1}},0xb046287d533b082d},Float64,1} to an object of type Float64                                                    
This may have arisen from a call to the constructor Float64(...),                                                    
since type constructors fall back to convert methods.                                                                
Stacktrace:
 [1] ode_interpolant! at /home/david/.julia/v0.6/OrdinaryDiffEq/src/dense/rosenbrock_interpolants.jl:33 [inlined]
```
However, using finite differences I get a solution, and the not in-place method even works with `ForwardDiff`.

ChrisRackauckas:
>Moreover, I shortly checked Rosenbrock23 again, and the in-place method fails because of some problem with ForwardDiff:

Peculiar. The Rosenbrocks and the SDIRKs should have the same Jacobian code, do they not?

ChrisRackauckas:
Oh, somethings wrong with the TimeDerivatives, which exist in Rosenbrock but not SDIRK. But there's a test which would catch that...?

devmotion:
I managed to link to existing caches. One has to take care of some more fields (not only derivative wrappers) since also `u`, `uprev`, `uprev2`, and e.g. fields defined as `uhold = vec(u)` have to be updated.
Moreover, I noticed that SDIRK caches currently have a needless parameter `tType` which can not inferred from the fields, and `UJacobianWrapper` and `JacobianConfig` have to be reordered to simplify updates of both fields. Hence this PR now requires https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/173 and depends on a yet to be released version 2.18.0 of OrdinaryDiffEq.

devmotion:
BTW, I figured out how to fix the Rosenbrock methods. In fact, it's a problem in OrdinaryDiffEq which only occurs when used in DelayDiffEq with in-place methods: the calculation of time derivatives causes (only in DelayDiffEq!) also via the history function the calculation of interpolants of the ODE solution, and only for in-place methods line
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/dense/generic_dense.jl#L255
then calls
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/dense/generic_dense.jl#L324
before dispatching to e.g.
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/dense/rosenbrock_interpolants.jl#L24.
However, this allocates output arrays in line
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/dense/generic_dense.jl#L326
which usually are correct - however, when used with `ForwardDiff` `Θ` and hence the result of the calculation is a dual whereas `y₀` contains regular numbers. Thus the type of the output array is incorrect which then causes an error in line
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/dense/rosenbrock_interpolants.jl#L33.
I could fix the problem by determining the output type beforehand:
```julia
@inline function ode_interpolant{TI}(Θ,dt,y₀,y₁,k,cache::OrdinaryDiffEqMutableCache,idxs,T::Type{Val{TI}})
  # determine output type
  # required for calculation of time derivatives with autodifferentiation
  oneunit_Θ = oneunit(Θ)
  S = promote_type(typeof(oneunit_Θ * oneunit(eltype(y₀))), # Θ*y₀
                   typeof(oneunit_Θ * oneunit(eltype(y₁))), # Θ*y₁
                   typeof(oneunit_Θ * oneunit(dt) * oneunit(eltype(k[1])))) # Θ*dt*k

  if typeof(idxs) <: Void
    out = similar(y₀, S)
  else
    !(typeof(idxs) <: Number) && (out = similar(y₀, S, indices(idxs)))
  end
  if typeof(idxs) <: Number
    return ode_interpolant!(nothing,Θ,dt,y₀,y₁,k,cache,idxs,T)
  else
    ode_interpolant!(out,Θ,dt,y₀,y₁,k,cache,idxs,T)
    return out
  end
end
```
Do you think there is a better fix, i.e. is there another way to completely circumvent that problem? Is it possible to directly use the in-place method for interpolants in the case of mutable caches?

ChrisRackauckas:
>Is it possible to directly use the in-place method for interpolants in the case of mutable caches?

It seems like that is happening in the user's history function. Then it would be up to them to provide the cache, which is even more of a hassle because they'd have to know how to make it a `DualCache` for it to handle ForwardDiff.jl. So I don't think we can rely on this.

>Do you think there is a better fix, i.e. is there another way to completely circumvent that problem?

No, and I think this is actually a good fix that makes the interpolant more robust anyways, while I do agree this is slightly messy and wish there was something easier. But I think this is necessary.

devmotion:
> No,  and I think this is actually a good fix that makes the interpolant more  robust anyways, while I do agree this is slightly messy and wish there  was something easier. But I think this is necessary.

Good, I'll create a PR 👍

ChrisRackauckas:
LGTM. I wish there was a way to make it more automatic, but you set it up so that extending the generated function isn't bad at all. Good job. 

[Though I think it won't be much of a problem since most of the architectures are "complete now": I'm not sure multistep stiff solvers or fully implicit RK methods will have very many extra ideas.]

ChrisRackauckas:
I'll do an OrdinaryDiffEq.jl release so that way we can get this tested.

test/sdirk_integrators.jl
ChrisRackauckas: Should Rosenbrocks get added now?

ChrisRackauckas:
https://github.com/JuliaLang/METADATA.jl/pull/10937

ChrisRackauckas:
Add Rosenbrocks to the test?

devmotion:
Can this be merged?

ChrisRackauckas:
Thanks so much! This is great to have!

PR #32, opened 2017-08-29
Do not return SubArrays in solution

devmotion:
This also fixes the problem described in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/180 for DelayDiffEq. Furthermore, interpolation and plotting seems to be broken for solutions that are returned as `SubArray` and I could not see an easy fix. Moreover, I fear that it might cause additional problems I am not aware of yet. So even though I would like to keep allocations down to a minimum, I guess we should rather not return `SubArrays`. Additionally, currently sometimes regular arrays and sometimes `SubArrays` are returned which seems inconsistent.

### DIFF
--- a/src/integrator_utils.jl
+++ b/src/integrator_utils.jl
@@ -86,15 +86,14 @@ function build_solution_array(integrator::DDEIntegrator)
             if typeof(integrator.opts.save_idxs) <: Void
                 u = integrator.sol.u
             else
-                u = [@view(u[integrator.opts.save_idxs]) for u in integrator.sol.u]
+                u = [u[integrator.opts.save_idxs] for u in integrator.sol.u]
             end
         else # remove initial time point
-            t = @view integrator.sol.t[2:end]
+            t = integrator.sol.t[2:end]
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Yes, I don't know why I didn't catch this before but there's also just no reason to use `view`s in the saving stuff because there's has to be a copy somewhere too, so you might as well do it when slicing. Just have to make sure that whatever happens further down doesn't accidentally copy. That's what the `Val{false}` is for when slicing here:

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/integrators/integrator_utils.jl#L94

with `copyat_or_push!` defined in RecursiveArrayTools.jl (like everything else lol).

devmotion:
I'm not sure where exactly it would be preferable to use `copyat_or_push!` - e.g. instead of the comprehensions?

ChrisRackauckas:
I see, this is a slightly different thing since the values are already saved. I think this is fine, at least for now.

PR #33, opened 2017-09-04
Update for DiffEqProblemLibrary changes

devmotion:
This updates tests to a possible next version of DiffEqProblemLibrary which does export problems instead of functions (see https://github.com/JuliaDiffEq/DiffEqProblemLibrary.jl/pull/13).

### DIFF
--- a/test/REQUIRE
+++ b/test/REQUIRE
@@ -1,4 +1,4 @@
 DiffEqDevTools
 DiffEqCallbacks
-DiffEqProblemLibrary
+DiffEqProblemLibrary 0.11.0
 Unitful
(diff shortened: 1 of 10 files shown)


devmotion:
@ChrisRackauckas how do you want to manage the update of DiffEqProblemLibrary? Do you want to add an upper bound to the current DiffEqProblemLibrary requirement first?

ChrisRackauckas:
I don't think we can add upper bounds to something that is only a test dependency. But lets get it merged and registered, giving a note to the METADATA maintainers that it will break tests for now.

ChrisRackauckas:
https://github.com/JuliaLang/METADATA.jl/pull/11064

PR #34, opened 2017-09-04
Add interpolants and change order of interpolation updates

devmotion:
This is a quite large update. Changes are:
- new discontinuities (arising from callbacks) are also added as time stops [this change alone causes the only regression in `events.jl`; just applying that change to master leads to the same regression]
- behaviour of ODE integrator is changed: currently ODE integrator and its interpolation data are only updated to the next interval `[t, t+dt]` during fixed-point iterations and directly before saving; with this PR it will always be updated to `[t, t+dt]` directly after calculating the next step and only be reset to the last interval `[tprev, t]` at the beginning of `perform_step!` if the last step failed. The advantage is that any evaluations of the history function (e.g. in callbacks when interpolating back) will already use the newest calculations; with the current approach they might again do a bad extrapolation as initially in `perform_steps!`.
- systematic updates of interpolation data: every time when the interpolation data of the ODE integrator is updated or moved to the next interval we instead update the current interpolation data or the one just calculated for the next interval of the DDE integrator and copy it then to the ODE integrator. The advantage is that then these updates will always use the old interpolation data of the ODE integrator when evaluating the history function. Currently updates of the first interpolation steps might change results for the last interpolation steps because the first updates already change the history function, so we can get undesired side effects. This can not happen when interpolation steps are added to the DDE integrator first (but we have to specify `tprev`, since `integrator.tprev != integrator.integrator.tprev` when moving the ODE integrator to the next interval!).
- constant extrapolation in the initial time step: currently extrapolation of the history function at the initial time step will fail since correctly `dt = 0` for the ODE integrator. This does not matter for constant delays since the initial time step is always chosen such that `dt < minimum(lag)` but this can not be ensured for dependent lags. Hence a special constant extrapolation of `integrator.u` is added to the history function in case `integrator.dt == 0`.
- initial interpolations of lazy interpolants: currently lazy interpolants fail if the time step is chosen such that in the calculation of additional interpolation steps the history function has to be evaluated at a point in the current time interval (this happens e.g. for the Mackey-Glass equation `prob_dde_mackey`, one of the new problems in DiffEqProblemLibrary). The evaluation will try to interpolate but has to add the additional steps first - so it can just not work. However, similar to the extrapolation that is necessary for implicit steps it will now fallback to a Hermite interpolant which then can be improved by further iterations using this first interpolation.
- better detection of implicit steps: instead of checking `dt > maximum(lag)` which will not work if not all constant lags are specified or additionally dependent lags exist it is now checked whether the history function was evaluated at a time point past the final time point of the solution, i.e. whether an extrapolation was used, during the initial calculation of the next step in `perform_step!`. This is done by setting `integrator.isout = true` which is always reset before this initial calculation. If an extrapolation occurred fixed-point iterations will be executed, otherwise not. This works for arbitrary delays and even without delays.
- update of "internals" after modifications by callbacks: a method `reeval_internals_due_to_modification!` for DDEIntegrator is added which will update the interpolation data of both the DDE integrator (again done first, using the "old" interpolation of the ODE integrator when evaluating the history function) and the ODE integrator (copy from DDE integrator), and the time interval of the ODE integrator.
- remove caches: since every step is saved reset of ODE integrators after failed steps is (at least partly) easy - we can just set it to the last time interval of the solution. The final element of the solution and its associated interpolation data will always be the values of the last time step and its interpolation data - however, this does not have to hold for the second-to-last element since callbacks can lead to additional savings. Hence the index of the `tprev` and `u(tprev)` in the solution arrays has to be saved and updated after successful steps. `dt` is still cached to avoid subtractions if possible.
- improved reduction of solution: it now also checks whether a time point of the solution is required for the interpolation at `tmax` up to which the solution should be reduced. However, it still does not work with callbacks, it will just remove the added steps again (maybe this can be fixed by just adding those additional points to `saveat` as well...).

### DIFF
--- a/src/history_function.jl
+++ b/src/history_function.jl
@@ -9,7 +9,7 @@ Before the initial time point of solution `sol` values are calculated by history
 and after the final time point of `sol` an inter- or extrapolation of the current state
 of integrator `integrator` is retrieved.
 """
-struct HistoryFunction{F1,F2,F3} <: Function
+struct HistoryFunction{F1,F2,F3<:ODEIntegrator} <: Function
     h::F1
     sol::F2
     integrator::F3
@@ -30,6 +30,9 @@ function (f::HistoryFunction)(t, deriv::Type=Val{0}, idxs=nothing)
 
     # set boolean that indicates that history function was evaluated at time point past the
     # final time point of the current solution
+    # NOTE: does not interfere with usual use of isout since this integrator is only used
+    # for inter- and extrapolation of future values and saving of the solution but does not
+    # affect whether time steps are accepted
     integrator.isout = true
 
     # handle extrapolations in initial time step (otherwise dt = 0 should not occur)
@@ -53,6 +56,9 @@ function (f::HistoryFunction)(val, t, deriv::Type=Val{0}, idxs=nothing)
 
     # set boolean that indicates that history function was evaluated at time point past the
     # final time point of the current solution
+    # NOTE: does not interfere with usual use of isout since this integrator is only used
+    # for inter- and extrapolation of future values and saving of the solution but does not
+    # affect whether time steps are accepted
     integrator.isout = true
 
     # handle extrapolations in initial time step (otherwise dt = 0 should not occur)
--- a/src/history_function.jl
+++ b/src/history_function.jl
@@ -24,8 +24,23 @@ function (f::HistoryFunction)(t, deriv::Type=Val{0}, idxs=nothing)
         end
     elseif t <= f.sol.t[end] # Put equals back
         return f.sol.interp(t, idxs, deriv)
+    end
+
+    integrator = f.integrator
+
+    # set boolean that indicates that history function was evaluated at time point past the
+    # final time point of the current solution
+    integrator.isout = true
+
+    # handle extrapolations in initial time step (otherwise dt = 0 should not occur)
+    integrator.dt == 0 && return constant_extrapolation(t, integrator, idxs, deriv)
+
+    if typeof(integrator.alg) <: Union{Vern6,Vern7,Vern8,Vern9} &&
+        length(integrator.k) == integrator.kshortsize
+        # use Hermite interpolant as fallback to avoid implicit calculation
+        return hermite_interpolant(t, integrator, idxs, deriv)
     else
-        return OrdinaryDiffEq.current_interpolant(t, f.integrator, idxs, deriv)
+        return OrdinaryDiffEq.current_interpolant(t, integrator, idxs, deriv)
     end
 end
 
@@ -38,6 +53,21 @@ function (f::HistoryFunction)(val, t, deriv::Type=Val{0}, idxs=nothing)
         end
     elseif t <= f.sol.t[end] # Put equals back
         return f.sol.interp(val, t, idxs, deriv)
+    end
+
+    integrator = f.integrator
+
+    # set boolean that indicates that history function was evaluated at time point past the
+    # final time point of the current solution
+    integrator.isout = true
+
+    # handle extrapolations in initial time step (otherwise dt = 0 should not occur)
+    integrator.dt == 0 && return constant_extrapolation!(val, t, integrator, idxs, deriv)
+
+    if typeof(integrator.alg) <: Union{Vern6,Vern7,Vern8,Vern9} &&
+        length(integrator.k) == integrator.kshortsize
+        # use Hermite interpolant as fallback to avoid implicit calculation
+        return hermite_interpolant!(val, t, integrator, idxs, deriv)
     else
         return OrdinaryDiffEq.current_interpolant!(val, t, f.integrator, idxs, deriv)
     end
--- a/src/history_function.jl
+++ b/src/history_function.jl
@@ -35,13 +35,7 @@ function (f::HistoryFunction)(t, deriv::Type=Val{0}, idxs=nothing)
     # handle extrapolations in initial time step (otherwise dt = 0 should not occur)
     integrator.dt == 0 && return constant_extrapolation(t, integrator, idxs, deriv)
 
-    if typeof(integrator.alg) <: Union{Vern6,Vern7,Vern8,Vern9} &&
-        length(integrator.k) == integrator.kshortsize
-        # use Hermite interpolant as fallback to avoid implicit calculation
-        return hermite_interpolant(t, integrator, idxs, deriv)
-    else
-        return OrdinaryDiffEq.current_interpolant(t, integrator, idxs, deriv)
-    end
+    return OrdinaryDiffEq.current_interpolant(t, integrator, idxs, deriv)
 end
 
 function (f::HistoryFunction)(val, t, deriv::Type=Val{0}, idxs=nothing)
@@ -64,11 +58,5 @@ function (f::HistoryFunction)(val, t, deriv::Type=Val{0}, idxs=nothing)
     # handle extrapolations in initial time step (otherwise dt = 0 should not occur)
     integrator.dt == 0 && return constant_extrapolation!(val, t, integrator, idxs, deriv)
 
-    if typeof(integrator.alg) <: Union{Vern6,Vern7,Vern8,Vern9} &&
-        length(integrator.k) == integrator.kshortsize
-        # use Hermite interpolant as fallback to avoid implicit calculation
-        return hermite_interpolant!(val, t, integrator, idxs, deriv)
-    else
-        return OrdinaryDiffEq.current_interpolant!(val, t, f.integrator, idxs, deriv)
-    end
+    return OrdinaryDiffEq.current_interpolant!(val, t, f.integrator, idxs, deriv)
 end
--- a/src/integrator_interface.jl
+++ b/src/integrator_interface.jl
@@ -4,26 +4,53 @@
 Update solution of `integrator`, if necessary or forced by `force_save`.
 """
 function savevalues!(integrator::DDEIntegrator, force_save=false)
-    # update ODE integrator
-    update_ode_integrator!(integrator)
+    # update time of ODE integrator (can be slightly modified (< 10ϵ) because of time stops)
+    if integrator.integrator.t != integrator.t
+        if abs(integrator.t - integrator.integrator.t) >= 10eps(integrator.t)
+            error("unexpected time discrepancy detected")
+        end
 
-    # update solution of ODE integrator
-    savevalues!(integrator.integrator, force_save)
-
-    if typeof(integrator.sol.prob) <: ConstantLagDDEProblem
-        #warn("ConstantLagDDEProblem is deprecated. Use DDEProblem instead.")
-        neutral = false
-        constant_lags = integrator.sol.prob.lags
-    else
-        neutral = integrator.sol.prob.neutral
-        constant_lags = integrator.sol.prob.constant_lags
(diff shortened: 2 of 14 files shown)


src/history_function.jl
```diff
@@ -24,8 +24,23 @@ function (f::HistoryFunction)(t, deriv::Type=Val{0}, idxs=nothing)
         end
     elseif t <= f.sol.t[end] # Put equals back
         return f.sol.interp(t, idxs, deriv)
+    end
+
+    integrator = f.integrator
+
+    # set boolean that indicates that history function was evaluated at time point past the
+    # final time point of the current solution
+    integrator.isout = true
```
ChrisRackauckas: This is usually for `isoutofdomain`. It is confusing to repurpose it like this. Add another boolean to the integrator instead since that's really cheap anyways.

src/integrator_interface.jl
```diff
@@ -4,26 +4,53 @@
...
     savevalues!(integrator.integrator, force_save)
 
-    if typeof(integrator.sol.prob) <: ConstantLagDDEProblem
-        #warn("ConstantLagDDEProblem is deprecated. Use DDEProblem instead.")
-        neutral = false
-        constant_lags = integrator.sol.prob.lags
-    else
-        neutral = integrator.sol.prob.neutral
-        constant_lags = integrator.sol.prob.constant_lags
+    # copy additional interpolation steps back from solution to reduced ODE integrator
```
ChrisRackauckas: I am not quite sure why this is necessary?

src/integrator_interface.jl
```diff
@@ -4,26 +4,53 @@
...
+    # last entry of solution always equals last time point of integration
+    if length(integrator.sol.k[end]) > length(integrator.integrator.k)
+        @inbounds for i in integrator.kshortsize+1:length(integrator.sol.k[end])
+            copyat_or_push!(integrator.integrator.k, i, integrator.sol.k[end][i])
+        end
+    end
+
+    # update prev2_idx to indices of tprev and u(tprev) in solution
```
ChrisRackauckas: what is this? I haven't seen this anywhere.

src/integrator_interface.jl
```diff
@@ -4,26 +4,53 @@
...
+    integrator.prev2_idx = integrator.prev_idx
+
+    # cache dt of interval [tprev, t] of ODE integrator since it can only be retrieved by
+    # a possibly incorrect subtraction
+    integrator.integrator.dtcache = integrator.integrator.dt
```
ChrisRackauckas: Will this make fixed timestep methods work incorrectly? `dtcache` already has a use.

src/integrator_interface.jl
```diff
@@ -48,77 +73,49 @@ end
...
-    OrdinaryDiffEq.ode_addsteps!(integrator.integrator, integrator.f)
+    # reset ODE integrator to cached values if last step failed
+    if !integrator.integrator.accept_step
+        if typeof(integrator.u) <: AbstractArray
+            recursivecopy!(integrator.integrator.u, integrator.sol.u[end])
+        else
+            integrator.integrator.u = integrator.sol.u[end]
+        end
+        integrator.integrator.t = integrator.sol.t[end]
+        integrator.integrator.tprev = integrator.sol.t[integrator.prev2_idx]
```
ChrisRackauckas: Oh I see what `prev2_idx` is.

src/integrator_interface.jl
```diff
@@ -48,77 +73,49 @@ end
...
-        if typeof(integrator.integrator.uprev) <: AbstractArray
-            recursivecopy!(integrator.integrator.uprev, integrator.uprev)
-        else
-            integrator.integrator.uprev = integrator.uprev
-        end
+    # if the history function was evaluated at time points past the final time point of the
+    # solution, i.e. returned extrapolated values, continue with a fixed-point iteration
+    if integrator.integrator.isout
+        # update ODE integrator to next time interval together with correct interpolation
+        advance_ode_integrator!(integrator)
```
ChrisRackauckas: Don't know what this is yet.

src/interpolants.jl
```diff
@@ -0,0 +1,97 @@
...
+Calculate Hermite interpolation of derivative `deriv` at time `t` and indices `idxs`
+for `integrator`.
+"""
+function hermite_interpolant(t, integrator::DEIntegrator, idxs, deriv)
+    [hermite_interpolant(τ, integrator, idxs, deriv) for τ in t]
+end
+
+function hermite_interpolant(t::Number, integrator::DEIntegrator, idxs, deriv)
+    Θ = (t - integrator.tprev)/integrator.dt
+    OrdinaryDiffEq.hermite_interpolant(Θ, integrator.dt, integrator.uprev,
```
ChrisRackauckas: This requires `k1 = f(t,uprev)` and `k2 = f(t+dt,u)`. Why would this be true for the methods which normally don't use Hermite?

test/residual_control.jl
```diff
@@ -6,30 +6,30 @@ u₀ = 1.0
...
+@test sol3.errors[:final] < 4.1e-9
+@test sol3.errors[:l2] < 9.2e-9
 
 sol4 = solve(prob2, alg, abstol=1e-13,reltol=1e-13)
 
-@test sol4.errors[:l∞] < 1e-1
-@test sol4.errors[:final] < 1e-3
-@test sol4.errors[:l2] < 1e-1
+@test sol4.errors[:l∞] < 1.2e-10
+@test sol4.errors[:final] < 2.0e-15
```
ChrisRackauckas: There's no way `2e-15` bounds are justified. Those are probably going to cause floating point issues.

test/unconstrained.jl
```diff
@@ -103,8 +103,8 @@ alg4 = MethodOfSteps(DP5(), constrained=false, max_fixedpoint_iters=100,
 sol4 = solve(prob, alg4)
 
 @test abs(sol1[end] - sol2[end]) < 2.5e-11
-@test abs(sol1[end] - sol3[end]) < 1.3e-13
-@test abs(sol1[end] - sol4[end]) < 4.5e-13
+@test abs(sol1[end] - sol3[end]) < 1.3e-14
```
ChrisRackauckas: why? won't this cause floating point issues?

src/history_function.jl
devmotion: I know, but it's an object of type `ODEIntegrator` that is never iterated and is only used for inter/extrapolation - so we would have to change it in OrdinaryDiffEq. I thought it's not worth it for just this special use case in DelayDiffEq. Shall I add a field to `ODEIntegrator` nevertheless?

src/integrator_interface.jl
devmotion: Because of line
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/integrators/integrator_utils.jl#L135
`integrator.integrator.k` is always reduced to a length of `integrator.kshortsize`.

src/integrator_interface.jl
devmotion: The `ODEIntegrator` is never iterated and besides inter/extrapolation only `savevalues!` and `postamble!` is applied to it. So I guess we can change everything that is not required for these tasks and methods without any effects - and hence we can also change `dtcache` arbitrarily.

src/interpolants.jl
devmotion: Oh I missed that this is necessary. Since it is absolutely necessary (in my opinion) to somehow calculate initial interpolants without adding additional interpolation steps (otherwise it is always possible that we end up with a recursion that will fail) we then have to check this for every such method (currently only Verner methods are handled) and then probably redefine own custom interpolants. In the case of Verner methods we could use `k1 = f(t, uprev)` but we then have to calculate `f(t+dt, u)` since this is different from all initial interpolation steps `ki`.

test/unconstrained.jl
devmotion: It didn't cause any issues on my computer and also not in the Travis builds (the failure there is due to a build error on Mac) and the Win32 build. Just two bounds in `residual_controls.jl` failed slightly, but in the range `1e-10` and `1e-11`, these here passed. I just wanted to make sure that any regressions or improvements are immediately visible in the tests. But I can just relax these bounds again.

test/residual_control.jl
devmotion: Why aren't they justified? Only two bounds of these and only on Win64 caused errors. Interestingly not the `2.0e-15` bound but two tests in the range of `1e-10` and `1e-11`.

src/integrator_interface.jl
devmotion: When the next step is calculated (at this time only once, it will continue with fixed-point iterations) we should add that information to the history function. Therefore the ODE integrator has already to move to the time interval `[t, t+dt]`, and this is done in `advance_ode_integrator!`. If the ODE integrator is not updated to the next time interval any call `h(t1)` for some `t1` in `[t, t+dt]` will still use the initial extrapolation of the ODE integrator from the last time step in `[tprev, t]`, so e.g. even in the next fixed-point iterations we would still use our initial guesses when evaluating the history function.

src/interpolants.jl
devmotion: I just rethought about this and it seems to me this whole problem is already solved by the changed order of interpolation updates of ODE integrator. I had added these special interpolants at the beginning of my update to the interpolation at a time when the order was different and there was no other way to get it running.
Just tested it with the Mackey-Glass equation which fails on master and apparently these special interpolants are not required anymore (at least I could not reproduce any errors). So even better, we can completely get rid of these special treatments and it should just work.

src/history_function.jl
ChrisRackauckas: I guess it should be fine. It doesn't interfere with `isoutofdomain` though?

src/integrator_interface.jl
ChrisRackauckas: Maybe we should add a `reduce_size` argument to `savevalues!` of `ODEIntegrator`? This is cheap so it's fine, but it's not the nicest way to handle it.

src/integrator_interface.jl
ChrisRackauckas: Comment it.

src/integrator_interface.jl
ChrisRackauckas: I see because it doesn't auto-advance anymore. Got it.

test/unconstrained.jl
ChrisRackauckas: I would just relax them a bit. A regression which shows something actually broke will make this many orders of magnitude larger, and I think they'll be small enough that JuliaCIBot or other things can have random breakage due to just calculating floating points differently.

devmotion:
The same here: needs https://github.com/JuliaLang/METADATA.jl/pull/11103

devmotion:
Need to merge https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/36 first, OrdinaryDiffEq 2.19.0 is currently incompatible with DelayDiffEq.

devmotion:
I'll close this since you approved and all commits are contained in https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/35, which is based on this PR.

ChrisRackauckas:
I think we should pull this in separately first.

devmotion:
I don't mind, I just thought it might be easier since we have to resolve conflicts in both PRs.

ChrisRackauckas:
I guess, but since this is such a large change I'd like to have a clean history.

PR #35, opened 2017-09-06
Discontinuity callback

devmotion:
This is an initial PR for adding support of dependent delays.

It builds upon https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/34 and adds both the type discontinuity and a callback that tracks discontinuities with dependent delays.

Moreover, I added a small example that compares a constant delay DDE, specified with constant delays, with just dependent delays (i.e. constant delays encoded as function), and without any delays. Unsurprisingly the last case leads to the worst results, but interestingly the best results are achieved with dependent delays. First I thought that failed steps caused by `DiscontinuityCallback` might still use some old and wrong cached values (and there might still be a problem here, I'm not sure whether the application of the discontinuity callback is done in a completely correct way), but the basic reason why these different delay specifications do not yield the same results is the fact that callbacks are not applied if a step is not accepted. Already in one of the first steps the integrator with dependent delays wants to step past the first discontinuity at `t = 1` - however, since this step just fails (error estimator too high) it is repeated with a smaller `dt` without application of the callback. The integrator with constant delays however wants to step to `t = 1`, i.e. wants to use a smaller step size already initially, and hence also the smaller step size after this step fails is below the step size in the case of dependent delays. Hence I guess we should not worry too much about these results being not completely the same.


src/DelayDiffEq.jl
```diff
@@ -5,34 +5,31 @@ module DelayDiffEq
...
 include("utils.jl")
 
-export init
-
-export initialize!, loopheader!, perform_step!, loopfooter!, handle_tstop!, postamble!
-
-export HistoryFunction
-
-export MethodOfSteps
+export Discontinuity, MethodOfSteps
```
ChrisRackauckas: Does Discontinuity have to be exported?

devmotion: Disontinuities have to be specified as vector of `Discontinuity` in `solve`, so yes. We could allow simple time vectors as well and cnvert those time points to discontinuities of order 0, but even then I think it is still good to be able to specify concrete discontinuities sometimes.

devmotion:
Another thing that came to my mind: currently `reset_fsal!` resets only the DDE integrator via the default implementation in OrdinaryDiffEq, i.e. in the calculation of the next step `integrator.fsalfirst = f(t, u)` instead of `integrator.fsalfirst = integrator.fsallast`. Does one also have to change/update the ODE integrator, which is responsible for extrapolating the history function, since according to Bellen one has to "restart [the ODE method] after each discontinuity point"? Would it be better to just use a constant extrapolation in the same way as it is done for the initial time step?

devmotion:
This makes `Discontinuity` a subtype of `Number`. I'm a bit unhappy that now `Unitful` is an additional dependency, just to avoid ambiguous `convert` methods...

ChrisRackauckas:
Why are the number subtype and conversions needed?

devmotion:
It's not needed per se, it's only needed if we do not add additional type conversions to `OrdinaryDiffEq`. As I mentioned in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/186#issuecomment-327775397, these conversions caused extreme regressions in some tests (although I still don't understand why). Probably we can work around that problem somehow, but I thought maybe it is even better to just let `Discontinuity` handle those conversions. Later I noticed that because of the `convert` methods in `Unitful` we need additional methods just for units... So I'm not sure yet whether we rather deal with that problem in `OrdinaryDiffEq` (and work on the corresponding PR) or in `DelayDiffEq`.

devmotion:
I removed the dependency on `Unitful` by removing the `Numbers` supertype. There's a quite minimal implementation of `Discontinuity` which contains ordering of discontinuities (required for the heap of discontinuities), ordering of discontinuities and numbers (required to check whether there exists a discontinuity at the current time point in `integrator_utils.jl`), and multiplication of discontinuities with numbers (required to check whether a discontinuity is contained in the time span of the problem in `solve.jl`). Tests pass locally with the current version of https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/186.

devmotion:
Also needs https://github.com/JuliaLang/METADATA.jl/pull/11103

devmotion:
Need to merge https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/36 first, OrdinaryDiffEq 2.19.0 is currently incompatible with DelayDiffEq.

ChrisRackauckas:
While waiting for Travis, do you think you can do a PR to docs which shows this all? The tutorial and solve pages should get updates.

ChrisRackauckas:
Needs conflicts resolved.

src/callbacks.jl
```diff
@@ -45,3 +32,202 @@ function reeval_internals_due_to_modification!(integrator::DDEIntegrator)
...
     integrator.u_modified = false
 end
+
+# track propagated discontinuities for dependent delays
+
+struct DiscontinuityCallback{F,D<:Discontinuity,A,R} <: AbstractContinuousCallback
+    lags::F
+    discontinuities::Vector{D}
+    interp_points::Int
+    atol::A
```
ChrisRackauckas: should stick to always using abstol and reltol in the context of diffeq.

src/callbacks.jl
```diff
@@ -45,3 +32,202 @@ function reeval_internals_due_to_modification!(integrator::DDEIntegrator)
...
+
+    # update interpolation data and calculate interpolation points
+    if callback.interp_points != 0
+      ode_addsteps!(integrator)
+    end
+    Θs = linspace(0, one(integrator.t), callback.interp_points)
+
+    # use temporary array of mutable caches in calculation of interpolants
+    if typeof(integrator.cache) <: OrdinaryDiffEq.OrdinaryDiffEqMutableCache
+        idxs_internal = eachindex(integrator.cache.tmp)
```
ChrisRackauckas: don't use this for indexing. Always `idxs_internal = nothing`.

src/integrator_interface.jl
```diff
@@ -5,13 +5,15 @@ Update solution of `integrator`, if necessary or forced by `force_save`.
 """
 function savevalues!(integrator::DDEIntegrator, force_save=false)
     # update time of ODE integrator (can be slightly modified (< 10ϵ) because of time stops)
-    if integrator.integrator.t != integrator.t
-        if abs(integrator.t - integrator.integrator.t) >= 10eps(integrator.t)
-            error("unexpected time discrepancy detected")
-        end
+    if typeof(integrator.t) <: AbstractFloat # does not work for units!
+        if integrator.integrator.t != integrator.t
+            if abs(integrator.t - integrator.integrator.t) >= 10eps(integrator.t)
```
ChrisRackauckas: Could we just multiply by `oneunit(integrator.t)`? `10eps(integrator.t)*oneunit(integrator.t)` for that to work?

test/dependent_delays.jl
```diff
@@ -0,0 +1,34 @@
...
+
+dde_int = init(prob_dde_1delay, alg)
+
+sol = solve!(dde_int)
+
+@test sol.errors[:l∞] < 3.7e-5
+@test sol.errors[:final] < 2.0e-5
+@test sol.errors[:l2] < 1.5e-5
+
+# constant delay specified as function
```
ChrisRackauckas: We should test this at lower tolerances too.

src/solve.jl
```diff
@@ -1,31 +1,65 @@
...
               dtmax= typeof(prob) <: ConstantLagDDEProblem ?
               (isempty(prob.lags) ? prob.tspan[2]-prob.tspan[1] : tType(7*minimum(prob.lags))) : isempty(prob.constant_lags) ?
               dtmax = prob.tspan[2]-prob.tspan[1] :
               tType(7*minimum(prob.constant_lags)),
               dt=zero(tType),
               saveat=tType[], save_idxs=nothing, save_everystep=isempty(saveat),
               save_start=true, dense=save_everystep && !(typeof(alg) <: Discrete),
-              minimal_solution=true, kwargs...) where
+              minimal_solution=true, discontinuity_interp_points::Int=10,
+              discontinuity_atol=tType(1//Int64(10)^12), discontinuity_rtol=0,
```
ChrisRackauckas: abstol reltol

src/integrator_utils.jl
```diff
@@ -275,3 +275,82 @@ function advance_ode_integrator!(integrator::DDEIntegrator)
...
+    if typeof(integrator.t) <: AbstractFloat # does not work for units!
+        while !isempty(integrator.opts.d_discontinuities) &&
+            abs(top(integrator.opts.d_discontinuities).t - integrator.t) < 10eps(integrator.t)
+
+            d2 = pop!(integrator.opts.d_discontinuities)
+            order = min(order, d2.order)
+        end
+
+        # also remove all corresponding time stops
+        while !isempty(integrator.opts.tstops) &&
```
ChrisRackauckas: won't this delete user-chosen timestops?

src/integrator_utils.jl
```diff
@@ -275,3 +275,82 @@ function advance_ode_integrator!(integrator::DDEIntegrator)
     # update prev_idx to index of t and u(t) in solution
     integrator.prev_idx = length(integrator.sol.t)
 end
+
```
ChrisRackauckas: Could you explain the strategy that's going on in this file? I am not getting it.

src/integrator_interface.jl
devmotion: No, we need something different `eps(integrator.t)` is not defined for units. So maybe something like `10eps(integrator.t/oneunit(t))*oneunit(integrator.t)`. The problem occurs in `OrdinaryDiffEq`, as well (in `integrator_utils.jl` e.g.).

ChrisRackauckas: I see. Can you make it separate?

src/callbacks.jl
devmotion: I'll change it 👍

src/callbacks.jl
devmotion: Ah, I see, I mixed something up when I copied and simplified `callbacks.jl` from OrdinaryDiffEq.

ChrisRackauckas: This was recently fixed.

src/integrator_utils.jl
devmotion: Yes, it will. I'm not sure whether this it necessary, but I thought it might be good to remove discontinuities that are too close to each other, which could happen e.g. with constant lags of `1/3` and `1`. But if we remove close discontinuities we should also remove close time stops, I thought. Of course, this does not differentiate between user-chosen and added time stops (such as newly found discontinuities) - I'm not sure whether this is a problem? We could also remove this completely in the beginning, and just add it if it seems useful for some concrete examples.

src/integrator_utils.jl
devmotion: The idea is: if we hit a discontinuity (this is checked in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/integrators/integrator_utils.jl#L349-L350), then we remove the discontinuity and maybe additional discontinuities at the current time point (and maybe also discontinuities and time stops that are close by...) in `handle_discontinuities!`. The order of the discontinuity at the current time point is defined as the lowest order of all these discontinuities. If the problem is not neutral, we will only add additional discontinuities (in `add_next_discontinuities!`) if this order is below the order of the algorithm (which in future should also depend on the used interpolants, I guess). If we add discontinuities, we just add discontinuities of the next order caused by constant lags (these we can calculate explicitly and just add them to `d_discontinuities` and `tstops`) and add the current discontinuity to `tracked_discontinuities` which is the array of old discontinuities that are checked by the `DiscontinuityCallback`.

src/integrator_interface.jl
devmotion: A separate PR for OrdinaryDiffEq with this change?

test/dependent_delays.jl
devmotion: I'm not completely sure what's missing... Just the same tests with lower tolerances, again with constant delays, dependent delays, and without delays?

ChrisRackauckas: Just dependent delays with the solver at lower tolerance, show that the errors get lower.

src/integrator_interface.jl
ChrisRackauckas: Yeah and here.

src/integrator_utils.jl
ChrisRackauckas: Oh I see. I guess it doesn't matter. For some reason I thought this could be re-deleted in the future, making it miss these timepoints. But I guess that won't happen.

src/integrator_utils.jl
ChrisRackauckas: Got it, I'll read with this in mind. Could this be commented?

devmotion: I tried to add comments to all steps of this algorithm, should this summary be added as well?

ChrisRackauckas: Yeah, I wasn't able to catch on a first read. I didn't get how things were aggregated together and why the minimum order, but when you lay it out like this it makes more sense.

ChrisRackauckas:
Accepting when the current comments are addressed.

src/integrator_utils.jl
devmotion: What do you mean by "re-deleted in the future"? The integrator won't stop at these future timepoints, but I thought this might be OK since they are so close.

ChrisRackauckas: Yeah, that's fine.

test/dependent_delays.jl
devmotion: OK, I added some additional tests at lower tolerances. Is this what you wanted to see?

ChrisRackauckas: Yeah looks good.

ChrisRackauckas:
Since there were some changes, merge when the next set of tests pass. Thanks! This is a big win!

ChrisRackauckas:
Cheers!

PR #36, opened 2017-09-07
Update DDEIntegrator

devmotion:
This updates `DDEIntegrator` to the newest commits of `OrdinaryDiffEq` in which `rate_prototype` was removed. Additionally I removed the field `prob` (also not existent in `ODEIntegrator`) since it always equals `sol.prob`.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,10 +1,9 @@
 julia 0.6
 DiffEqBase 1.21.0
-OrdinaryDiffEq 2.18.0
+OrdinaryDiffEq 2.19.0
 DataStructures 0.4.6
 RecursiveArrayTools 0.2.0
 Combinatorics
-Compat 0.17.0
 Reexport
 MuladdMacro
 ForwardDiff
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -15,5 +15,5 @@
 
     push!(integrator.opts.d_discontinuities,
           compute_discontinuity_tree(constant_lags, integrator.alg,
-                                     integrator.t,integrator.prob.tspan[2],neutral)...)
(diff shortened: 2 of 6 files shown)


ChrisRackauckas:
Probably need to tag OrdinaryDiffEq only after there's an upper bound on DelayDiffEq then.

devmotion:
No, this PR does not work with the current release of `OrdinaryDiffEq` and the current master of `DelayDiffEq` does not work with the current master of `OrdinaryDiffEq`.

devmotion:
Can I just add an upper bound of 2.18.0 to the DelayDiffEq versions 0.9, 0.9.1, and 0.10.0 in https://github.com/JuliaLang/METADATA.jl/tree/metadata-v2/DelayDiffEq/versions and open a PR? Version 0.8.0  already contains an upper bound of 2.15.0.

ChrisRackauckas:
Leave previous upper bounds in place and place upper bounds on all of the newest versions. Yup, it's just a PR to METADATA.

devmotion:
Needs OrdinaryDiffEq 2.19.0, just noticed I forgot to update `REQUIRE`

devmotion:
Apropos `REQUIRE`: I know you readded `DiffEqBase` some time ago, but why is that required? I thought it's already covered since `OrdinaryDiffEq` requires it, and `DelayDiffEq` requires `OrdinaryDiffEq`. Are requirements not transitive?

ChrisRackauckas:
>I thought it's already covered since OrdinaryDiffEq requires it, and DelayDiffEq requires OrdinaryDiffEq. Are requirements not transitive?

Technically yes, but you shouldn't rely on it. This packages directly uses things that are from DiffEqBase.jl and so it should require DiffEqBase.jl

devmotion:
Ah I see.

PR #37, opened 2017-09-08
Relax bound that fails on Win32

devmotion:
The latest commit fails on Win32 since one bound in `unconstrained.jl` seems to be too strict (`2.51e-11 < 2.5e-11`). I was wondering why this error was introduced, and figured out that it is caused by line

https://github.com/JuliaDiffEq/DiffEqProblemLibrary.jl/blob/master/src/dde_premade_problems.jl#L347

which changed the lags of the affected problem from `[1//3, 1//5]` to `[1/3, 1/5]`. Manually creating this problem with old lags fixes this issue, however I thought it might be reasonable to just relax the affected bound. The lags were changed in `DiffEqProblemLibrary` to allow use of these problems with units (as e.g. done in `units.jl`).

The second change just removes a duplicate method.


ChrisRackauckas:
Merge when tests pass.

Issue #40, opened 2017-10-04
MethodError at early times


pwstokes225:
I receive a MethodError when trying to solve the following DDE problem:
`DDEProblem((t,u,h,du) -> du[1] = -h(t/2)[1], t -> [1.0], [1.0], (0.0, 10.0), [])`

The error only occurs if tspan starts sufficiently early.
e.g. It works when tspan = (0.1,10.0), but not when tspan = (0.001,10.0).

devmotion:
Thanks for bringing this up! Extrapolations at the initial time point should have been handled by https://github.com/JuliaDiffEq/DelayDiffEq.jl/blob/master/src/history_function.jl#L39, but apparently this does not work (at least not when calculating the initial time step...). Moreover, there is a type in this line (`constant_extrapolation` should rather be `constant_extrapolant`).

I fixed these issues in branch https://github.com/devmotion/DelayDiffEq.jl/tree/fix_constant_extrapolation. With these changes I could successfully solve your DDE problem.

pwstokes225:
I can also confirm this has fixed it. Thanks!

PR #42, opened 2017-10-04
Fix constant extrapolation

devmotion:
Constant extrapolation at the initial time point currently does not work, as shown in https://github.com/JuliaDiffEq/DelayDiffEq.jl/issues/40.

We should also add tests for `HistoryFunction` which should have catched these errors.

### DIFF
--- a/src/history_function.jl
+++ b/src/history_function.jl
@@ -36,9 +36,11 @@ function (f::HistoryFunction)(t, deriv::Type=Val{0}, idxs=nothing)
     integrator.isout = true
 
     # handle extrapolations at initial time point
-    f.sol.t[end] == integrator.sol.prob.tspan[1] && return constant_extrapolant(t, integrator, idxs, deriv)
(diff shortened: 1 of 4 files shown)


ChrisRackauckas:
How about adding that test problem?

devmotion:
I added the test problem, but also some other more basic tests which should cover most functionality of `HistoryFunction`. These tests catched two additional bugs.

ChrisRackauckas:
For reference what were those bugs?

devmotion:
Two bugs in the implementation of `constant_extrapolant!`: a typo in line https://github.com/JuliaDiffEq/DelayDiffEq.jl/blob/master/src/interpolants.jl#L47 (should be `val` instead of `out`) and a missing `[idxs]` in line https://github.com/JuliaDiffEq/DelayDiffEq.jl/blob/master/src/interpolants.jl#L61.

PR #45, opened 2017-10-17
allow state changes from initialize

ChrisRackauckas:
https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/194

Mostly mirroring https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/213

I'm not sure this is totally correct though since it may need to change some `integrator.integrator` values?

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -55,7 +55,8 @@ function init(prob::AbstractDDEProblem{uType,tType,lType,isinplace}, alg::algTyp
                                                        tracked_discontinuities,
                                                        discontinuity_interp_points,
                                                        discontinuity_abstol,
-                                                       discontinuity_reltol)
+                                                       discontinuity_reltol,
+                                                       initialize!)
         callbacks = CallbackSet(callback, discontinuity_callback)
     else
         callbacks = callback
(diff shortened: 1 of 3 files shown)


devmotion:
I'm not sure... Shouldn't we do something similar to `handle_callback_modifiers!` and `reeval_internals_due_to_modification!` in the beginning of file https://github.com/JuliaDiffEq/DelayDiffEq.jl/blob/master/src/callbacks.jl? At least it seems reasonable to add a discontinuity of order 0 at the initial point (which might have been of order 1 before).

Moreover, we should probably replace most of the `integrator` with `dde_int` - `integrator` is just the ODE integrator that was used to construct `dde_int` and is responsible for interpolation.

ChrisRackauckas:
We don't want to `add_steps!` though. Maybe we should add a trigger in there to avoid `add_steps!` specifically via an optional arg.

src/solve.jl
```diff
@@ -296,8 +296,35 @@ function init(prob::AbstractDDEProblem{uType,tType,lType,isinplace}, alg::algTyp
 
     # set up additional initial values of newly created DDE integrator
     # (such as fsalfirst) and its callbacks
-    initialize!(dde_int)
-    initialize!(integrator.opts.callback, integrator.t, u, dde_int)
+
+    if initialize_integrator
```
devmotion: According to the failed tests it has to be added to list of keywords, as in OrdinaryDiffEq.

src/solve.jl
```diff
@@ -296,8 +296,35 @@ function init(prob::AbstractDDEProblem{uType,tType,lType,isinplace}, alg::algTyp
...
+      integrator.u_modified = true
```
devmotion: I think we should (mostly) only operate on `dde_int` here and in the following lines. `integrator` is the wrong integrator, callbacks are initialized with `dde_int`.

ChrisRackauckas:
Should be good now. @devmotion could you take a look?

devmotion:
LGTM. With the recently tagged DiffEqBase version tests now also pass on Travis.

ChrisRackauckas:
oh I see what happened to the tests. The fact that Travis Mac takes forever actually worked in our favor there lol.

PR #49, opened 2017-11-03
add reinit

ChrisRackauckas:


ChrisRackauckas:
The test currently fails because it steps to a different value than the first time at the third step for reasons unknown.

devmotion:
I had only a quick look but I guess we can't just use `OrdinaryDiffEq.tstop_saveat_disc_handling` since in `solve!` discontinuities are derived differently in lines https://github.com/JuliaDiffEq/DelayDiffEq.jl/blob/d4b4595f7848a235175e0e6c2b95ad9dd847fecd/src/solve.jl#L21-L38 (the order of the discontinuity at the initial time point might also change if `u` is modified). Moreover, `tracked_discontinuities` should probably be reset as well.

PR #59, opened 2018-03-11
Fix tests

devmotion:
This PR fixes an issue with units that is caused by https://github.com/ajkeller34/Unitful.jl/commit/a76397c75dcfa59d8328ab60effb55a9ad4d0c6a. Moreover, I had to adjust two error bounds for the tests to pass - were there any changes to RK4 that might have caused these test failures?


ChrisRackauckas:
I wouldn't worry about the tolerance change. They are so low that I think the fact that `@fastmath` or a single operation is re-ordered could do it. There hasn't been a change to `RK4` but I think there was a small timestep choice change (`@fastmath` and a small flip), so.

PR #60, opened 2018-03-15
Restructure tests and support purely in-place history

devmotion:
This is a huge PR but it mainly just restructures the existing tests. Similar to the tests of [NLsolve.jl](https://github.com/JuliaNLSolvers/NLsolve.jl), `@testset`s are now defined in the included files instead of `runtests.jl` and tests are better structured by nested `@testset`s and for loops of `@testset`s. The new structure helps with defining methods only locally; before in some cases the definition of a new function only added a new method to an already existing function. In particular, because of this problem the tests succeeded in the case of in-place history functions since additionally some out-of-place methods existed - although the `init` method required an out-of-place history function or an explicit `initial_order` keyword. Hence in addition to the mentioned refactoring of the tests this PR changes the default value of `initial_order` using a newly defined method `agrees` (I'm not quite satisfied with this name...) such that also purely in-place history functions are supported and the restructured and improved tests do not fail.

Since this PR is quite big already, I did not add any further implementation changes. However, I guess one should also require the history function to include parameters, i.e. users should define functions `h(p, t[, ::Type{Val{deriv}}; idxs])` or `h(val, p, t[, ::Type{Val{deriv}}; idxs])`. This would enable a simple way of dealing with parametrized history functions (and thus completely parametrized DDEs, it seems),  and I think, it should be easy to implement and improve consistency throughout the DiffEq ecosystem.

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -9,7 +9,7 @@ function init(prob::AbstractDDEProblem{uType,tType,lType,isinplace}, alg::algTyp
               dense=save_everystep && !(typeof(alg) <: FunctionMap),
               minimal_solution=true, discontinuity_interp_points::Int=10,
               discontinuity_abstol=tType(1//Int64(10)^12), discontinuity_reltol=0,
-              initial_order=prob.h(prob.tspan[1]) == prob.u0 ? 1 : 0,
+              initial_order=agrees(prob.h, prob.u0, prob.tspan[1]) ? 1 : 0,
               initialize_integrator = true, initialize_save = true,
               callback=nothing, kwargs...) where
     {uType,tType,lType,isinplace,algType<:AbstractMethodOfStepsAlgorithm}
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -1,4 +1,26 @@
 """
+    agrees(h, u, t)
+
+Determine whether history function evaluates to u at time point t.
+"""
+function agrees(h, u, t)
+    # Obtain signatures of h
+    sigs = [m.sig for m in methods(h)]
+
+    # Compare evaluation of h at time point t with u
+    if any(sig<:Tuple{Any, Any} for sig in sigs)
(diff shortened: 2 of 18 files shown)


devmotion:
@ChrisRackauckas Any comments? Tests are successful, and coverage increases even. Do you have any objections or can I merge?

ChrisRackauckas:
Let me take some time to review it, but I will likely merge it in the next day.

ChrisRackauckas:
>Since this PR is quite big already, I did not add any further implementation changes. However, I guess one should also require the history function to include parameters, i.e. users should define functions h(p, t[, ::Type{Val{deriv}}; idxs]) or h(val, p, t[, ::Type{Val{deriv}}; idxs]). This would enable a simple way of dealing with parametrized history functions (and thus completely parametrized DDEs, it seems), and I think, it should be easy to implement and improve consistency throughout the DiffEq ecosystem.

Yes, that would be a very good change, but I agree it should be a next PR.

src/solve.jl
```diff
@@ -9,7 +9,7 @@ function init(prob::AbstractDDEProblem{uType,tType,lType,isinplace}, alg::algTyp
               dense=save_everystep && !(typeof(alg) <: FunctionMap),
               minimal_solution=true, discontinuity_interp_points::Int=10,
               discontinuity_abstol=tType(1//Int64(10)^12), discontinuity_reltol=0,
-              initial_order=prob.h(prob.tspan[1]) == prob.u0 ? 1 : 0,
+              initial_order=agrees(prob.h, prob.u0, prob.tspan[1]) ? 1 : 0,
```
ChrisRackauckas: That's all that was necessary for purely inplace history functions?

devmotion: Yes, this (and the definition of `agrees`, of course) was the only change that was needed.

PR #61, opened 2018-03-25
Add parameters to history function

devmotion:
This PR adds parameters to the history function by default, i.e. with this change history functions have to be provided by the user in the forms `h(p, t[; idxs=nothing])` or `h(val, p, t[; idxs=nothing])`, and `h(p, t, Val{deriv}[; idxs=nothing])` or `h(val, p, t, Val{deriv}[; idxs=nothing])` for higher-order derivatives. Moreover, `reinit!` is fixed and now also works with in-place history functions (I missed this bug in https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/60 and added tests for it).

Currently tests fail since this PR requires a new release of DiffEqProblemLibrary with DDE problems updated to the new interface.

### DIFF
--- a/src/history_function.jl
+++ b/src/history_function.jl
@@ -15,16 +15,16 @@ struct HistoryFunction{F1,F2,F3<:ODEIntegrator} <: Function
     integrator::F3
 end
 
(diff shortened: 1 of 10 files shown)


ChrisRackauckas:
I'm going to do a major release with this. We'll need matching docs changes.

PR #65, opened 2018-04-12
update config building

ChrisRackauckas:
It should get setup for finite difference as well

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,6 @@
 julia 0.6
 DiffEqBase 3.0.3
-OrdinaryDiffEq 3.7.0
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Rosenbrock uses `du2` for what SDIRK uses `dz`. Not sure of a good way to fix.

devmotion:
You can dispatch on the cache, it's the third argument: https://github.com/JuliaDiffEq/DelayDiffEq.jl/blob/master/src/utils.jl#L52-L59
That would solve the problem, I guess?

ChrisRackauckas:
Yup, I can do that.

PR #66, opened 2018-04-14
Update for DiffEqBase.check_error!

devmotion:
Current master/next release of OrdinaryDiffEq breaks DelayDiffEq since `@ode_exit_conditions` requires `DiffEqBase.check_error!`. I guess the latest release OrdinaryDiffEq 3.10.0 should be added as an upper bound to previous versions of DelayDiffEq on METADATA.jl as well?

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,6 @@
 julia 0.6
-DiffEqBase 3.0.3
-OrdinaryDiffEq 3.10.0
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
I'm trying to avoid upper bounds because the package resolver is all screwed up, but I'll just put these in together.

PR #72, opened 2018-07-14
0.7 compatibility

YingboMa:


ChrisRackauckas:
Generic SDIRKs fail because NLsolve needs a tag, but that's it. Merging.

PR #76, opened 2018-07-17
Fix deprecations

devmotion:
Fixes some deprecations.

By the way, is the massive regression in both time and allocations known (compare e.g. https://travis-ci.org/JuliaDiffEq/DelayDiffEq.jl/jobs/402372843 with https://travis-ci.org/JuliaDiffEq/DelayDiffEq.jl/jobs/404760135)?


ChrisRackauckas:
>By the way, is the massive regression in both time and allocations known

Depwarns are extremely slow. Like orders of magnitude. I wouldn't trust any timings until they are all gone.

devmotion:
I see, good to know :+1:

PR #83, opened 2018-10-15
Fix `save_on`

YingboMa:
It seems that `save_on` is passed to `DEOptions` in a wrong order. Per https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/solve.jl#L259-L260 and https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/integrators/type.jl#L35-L36, `save_on` should be passed immediately after `dense`.


MSeeker1340:
Ah I was just looking at fixing DDE, thanks for the work.

Why the unexpected passes though? These should not have anything to do with this PR... Maybe new releases of OrdinaryDiffEq somehow fixed them?

devmotion:
These tests already passed on Julia nightly 14 days ago: https://travis-ci.org/JuliaDiffEq/DelayDiffEq.jl/jobs/435970728 So it might be due to some changes in Julia 1.0.1 rather than to changes in OrdinaryDiffEq, I guess.

devmotion:
Interestingly the tests only pass with Julia 1.0.1 and fail on the current Julia nightly.

ChrisRackauckas:
I'm not sure why we're getting those floating point differences now in v1.0. The SSA passes must've introduced a lot more optimizations on scalar code, making it heavily different from the vector code. We really need to drill down into this a bit after the newer APIs stabilize.

PR #85, opened 2018-10-17
Unify callback handling

YingboMa:
Ref: https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/164

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,6 @@
 julia 1.0
-DiffEqBase 3.10.0
-OrdinaryDiffEq 4.28.0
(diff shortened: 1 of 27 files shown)


ChrisRackauckas:
Looks like this will need a lower bound on OrdinaryDiffEq.jl as well.

ChrisRackauckas:
Well we need to merge this fix because otherwise everything is broken, but unit handling isn't something we tested on this. We probably want to apply `internalnorm` to the callback value to make it unitless for the storage, and then the typing is easier.

Issue #82, opened 2018-10-05
Instability when integrating backwards


pwstokes225:
The following bails with an instability warning:
```
using DelayDiffEq
f(du,u,h,p,t) = du .= h(p,t+1)
sol = solve(DDEProblem(f,[1.0],(p,t)->[1.0],(1.0,0.0)),MethodOfSteps(RK4()))
```

The solution expected here is the straight line u(t) = t.

ChrisRackauckas:
Thanks for the report. Likely `tdir` isn't getting passed down correctly.

devmotion:
I got this example working on my computer, the main issue were lines like https://github.com/JuliaDiffEq/DelayDiffEq.jl/blob/master/src/history_function.jl#L19 which do not respect `tdir`. I'm too tired right now but I'll add some tests and open a PR in the next days.

PR #88, opened 2018-11-04
Fix backward integration

devmotion:
Fixes https://github.com/JuliaDiffEq/DelayDiffEq.jl/issues/82.

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -38,6 +38,14 @@ function DiffEqBase.__init(
     integrator = init(ode_prob, alg.alg; initialize_integrator=false,
                       dt=one(tType), dtmax=dtmax, kwargs...)
 
+    # check that constant lags match the given time direction
+    if constant_lags != nothing
+        for lag in constant_lags
+            integrator.tdir * lag > zero(tType) ||
+                error("Constant lags and time direction do not match. Exiting.")
+        end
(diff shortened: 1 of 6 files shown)


src/solve.jl
```diff
@@ -261,6 +269,16 @@ function DiffEqBase.__init(
...
+        if sign(dde_int.dt)!=dde_int.tdir && dde_int.dt!=tType(0) && !isnan(dde_int.dt)
+            error("Automatic dt setting has the wrong sign. Exiting. Please report this error.")
+        end
+        if isnan(dde_int.dt)
+            if verbose
+                @warn("Automatic dt set the starting dt as NaN, causing instability.")
+            end
+        end
+    elseif dde_int.opts.adaptive && dde_int.dt > zero(dde_int.dt) && dde_int.tdir < 0
+        dde_int.dt *= dde_int.tdir # Allow positive dt, but auto-convert
```
devmotion: This is just a copy of https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/solve.jl#L328-L340, so it could be simplified e.g. by defining this code block with `@def` in OrdinaryDiffEq.

ChrisRackauckas: We could make this a function.

PR #90, opened 2018-12-29
make integrator type {Alg, IIP, U, T}

Datseris:


devmotion:
The inner constructor https://github.com/JuliaDiffEq/DelayDiffEq.jl/blob/f52b0f8f697469cca2cfdeb8c1ccf56e9edf7cef/src/integrator_type.jl#L61-L84 and the auxiliary function [`fsal_typeof`](https://github.com/JuliaDiffEq/DelayDiffEq.jl/blob/master/src/utils.jl#L29-L37) should be updated as well.

src/integrator_type.jl
```diff
@@ -55,7 +58,7 @@ mutable struct DDEIntegrator{algType<:OrdinaryDiffEq.OrdinaryDiffEqAlgorithm,
     fsallast::FSALType
 
     # incomplete initialization without fsalfirst and fsallast
-    function DDEIntegrator{algType,uType,tType,P,eigenType,absType,relType,
+    function DDEIntegrator{alg, IIP, uType, tType, P,eigenType,absType,relType,
```
devmotion: The `where` clause has to be updated as well and I guess we have to add the type parameters to `new` as well since `IIP` can not be inferred otherwise - but maybe I'm wrong and the correct signature pf the inner constructor is sufficient?

By the way, I'm wondering whether we should add a constructor that deduces IIP from the given DDEFunction?

Issue #93, opened 2019-01-18
saveat doubles the end point


ChrisRackauckas:
```julia
function delay_lotka_volterra(du,u,h,p,t)
  x, y = u
  α, β, δ, γ = p
  du[1] = dx = (α - β*y)*h(p,t-0.1)[1]
  du[2] = dy = (δ*x - γ)*y
end
h(p,t) = ones(eltype(p),2)
prob = DDEProblem(delay_lotka_volterra,[1.0,1.0],h,(0.0,10.0),[2.2, 1.0, 2.0, 0.4],constant_lags=[0.1])
sol = solve(prob,MethodOfSteps(Tsit5()),saveat=0.1)
length(sol)
length(0:0.1:10.0)
```

devmotion:
An easy fix is to make the inequality in https://github.com/JuliaDiffEq/DelayDiffEq.jl/blob/master/src/integrator_utils.jl#L99 strict; in this example the final time point is included in `saveat_internal` due to https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/0196b4fe34cce662d7069242f6f8668271c3fb98/src/solve.jl#L396 but the final time point is added anyway in the end, see https://github.com/JuliaDiffEq/DelayDiffEq.jl/blob/master/src/integrator_utils.jl#L180-L185.

devmotion:
Nevertheless, IMO the logic of how solutions are created should be simplified if possible... It feels a bit overcomplicated at the moment :smile:

ChrisRackauckas:
Indeed, it's quite complicated.

PR #94, opened 2019-01-18
Avoid duplicate end points with saveat

devmotion:
Fix https://github.com/JuliaDiffEq/DelayDiffEq.jl/issues/93.

### DIFF
--- a/src/integrator_utils.jl
+++ b/src/integrator_utils.jl
@@ -96,7 +96,7 @@ function build_solution_array(integrator::DDEIntegrator)
     else
         # calculate number of additional time points of solution
         saveat_length = count(
-            x -> integrator.tdir * x ≤ integrator.tdir * integrator.sol.t[end],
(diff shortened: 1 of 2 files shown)


devmotion:
Any idea for why Vern6 and Vern9 tests fail on Julia stable (in my branch) and on Julia nightlies (in your branch) but succeed on nightlies and on stable, respectively? Seems strange....

devmotion:
The same error occurred before: https://travis-ci.org/JuliaDiffEq/DelayDiffEq.jl/jobs/475655707

ChrisRackauckas:
It might be a small floating point difference that causes a rejected step to change to accepted in the different forms.

ChrisRackauckas:
Since it's working on the nightly I'll merge it, but it is quite curious.

devmotion:
I'm especially surprised about the fact that it both fails and succeeds on the nightly and the stable version from time to time - maybe it just depends on which computer/architecture is used for the Travis tests?

PR #95, opened 2019-01-20
Change initialization of tolerances of fixed point iterations

devmotion:
This PR changes how the tolerances of fixed point iterations are initialized. If unspecified, the error tolerances of the integrator are used without any conversions - the correct types are supposed to be set by the logic in OrdinaryDiffEq (https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/solve.jl#L112-L134). If they are specified, then it is only ensured that they are real numbers (similar to https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/solve.jl#L121 and https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/solve.jl#L133). Hence if units are used and tolerances are specified they always have to be given with correct units. Currently unitless numbers were converted to the correct units but in my opinion we shouldn't do that since it is not transparent to the user.

This PR, however, has bigger implications: when used with Flux currently `fixedpoint_reltol_internal` is initialized as `TrackedReal{TrackedReal{Float64}}` which is incorrect and causes a stack overflow. With this PR the following code (modification of https://github.com/JuliaLang/www.julialang.org/pull/274) runs without any issues:
```julia
using DelayDiffEq
using DiffEqFlux
using Flux
using Plots

gr()

function delay_lotka_volterra(du,u,h,p,t)
  x, y = u
  α, β, δ, γ = p
  du[1] = dx = (α - β*y)*h(p,t-0.1)[1]
  du[2] = dy = (δ*x - γ)*y
end

h(p,t) = ones(eltype(p),2)

prob = DDEProblem(delay_lotka_volterra,[1.0,1.0],h,(0.0,10.0),constant_lags=[0.1])

p = param([2.2, 1.0, 2.0, 0.4])
params = Flux.Params([p])

reduction(sol) = sol[1,:]

function predict_rd_dde()
  diffeq_rd(p,reduction,prob,MethodOfSteps(Tsit5()),saveat=0.1)
end

loss_rd_dde() = sum(abs2,x-1 for x in predict_rd_dde())

data = Iterators.repeated((), 100)
opt = ADAM(0.1)
cb = function ()
  display(loss_rd_dde())
  display(plot(solve(remake(prob,p=Flux.data(p)),MethodOfSteps(Tsit5()),saveat=0.1),
          ylim=(0,6)))
end

cb()

Flux.train!(loss_rd_dde, params, data, opt, cb = cb)
```

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -92,12 +92,11 @@ function DiffEqBase.__init(
                             mass_matrix = prob.f.mass_matrix)
     end
 
-    # absolut tolerance for fixed-point iterations has to be of same type as elements of u
-    # in particular important for calculations with units
-    if typeof(alg.fixedpoint_abstol) <: Nothing
-        fixedpoint_abstol_internal = map(eltype(uType), integrator.opts.abstol)
+    # define absolut tolerance for fixed-point iterations
+    if alg.fixedpoint_abstol === nothing
+        fixedpoint_abstol_internal = integrator.opts.abstol
     else
-        fixedpoint_abstol_internal = map(eltype(uType), alg.fixedpoint_abstol)
+        fixedpoint_abstol_internal = real.(alg.fixedpoint_abstol)
     end
 
     # use norm of ODE integrator if no norm for fixed-point iterations is specified
@@ -105,18 +104,16 @@ function DiffEqBase.__init(
         fixedpoint_norm = integrator.opts.internalnorm
     end
 
-    # derive unitless types
-    uEltypeNoUnits = typeof(recursive_one(integrator.u))
-    tTypeNoUnits = typeof(recursive_one(prob.tspan[1]))
-
-    # relative tolerance for fixed-point iterations has to be of same type as elements of u
-    # without units
-    # in particular important for calculations with units
-    if typeof(alg.fixedpoint_reltol) <: Nothing
-        fixedpoint_reltol_internal = convert.(eltype(uType), integrator.opts.reltol)
+    # define relative tolerance for fixed-point iterations
+    if alg.fixedpoint_reltol === nothing
+        fixedpoint_reltol_internal = integrator.opts.reltol
     else
-        fixedpoint_reltol_internal = convert.(eltype(uType), oneunit.(integrator.u).*alg.fixedpoint_reltol)
+        fixedpoint_reltol_internal = real.(alg.fixedpoint_reltol)
     end
+
+    # derive unitless types
+    uEltypeNoUnits = recursive_unitless_eltype(prob.u0)
+    tTypeNoUnits = typeof(one(tType))
 
     # create separate copies u and uprev, not pointing to integrator.u or integrator.uprev,
     # containers for residuals and to cache uprev with correct dimensions and types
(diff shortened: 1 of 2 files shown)


src/solve.jl
```diff
@@ -92,32 +92,29 @@ function DiffEqBase.__init(
...
     end
 
-    # absolut tolerance for fixed-point iterations has to be of same type as elements of u
-    # in particular important for calculations with units
-    if typeof(alg.fixedpoint_abstol) <: Nothing
-        fixedpoint_abstol_internal = map(eltype(uType), integrator.opts.abstol)
+    # define absolut tolerance for fixed-point iterations
+    if alg.fixedpoint_abstol === nothing
+        fixedpoint_abstol_internal = integrator.opts.abstol
     else
```
devmotion: I guess we should copy abstol.

src/solve.jl
```diff
@@ -92,32 +92,29 @@ function DiffEqBase.__init(
...
-    tTypeNoUnits = typeof(recursive_one(prob.tspan[1]))
-
-    # relative tolerance for fixed-point iterations has to be of same type as elements of u
-    # without units
-    # in particular important for calculations with units
-    if typeof(alg.fixedpoint_reltol) <: Nothing
-        fixedpoint_reltol_internal = convert.(eltype(uType), integrator.opts.reltol)
+    # define relative tolerance for fixed-point iterations
+    if alg.fixedpoint_reltol === nothing
+        fixedpoint_reltol_internal = integrator.opts.reltol
```
devmotion: Same here.

src/solve.jl
ChrisRackauckas: why copy?

devmotion: I was thinking about use cases were you want to adjust the error tolerances of the integrator adaptively and I guess it's not (always) intended that the tolerances of the fixed point iterations change in the same way.

src/solve.jl
```diff
@@ -94,7 +94,12 @@ function DiffEqBase.__init(
 
     # define absolute tolerance for fixed-point iterations
     if alg.fixedpoint_abstol === nothing
-        fixedpoint_abstol_internal = integrator.opts.abstol
+        abstol = integrator.opts.abstol
+        if typeof(abstol) <: Number
+            fixedpoint_abstol_internal = abstol
+        else
```
devmotion: I'm not sure whether this specialization is required. To me it seems `recursivecopy(::Number)` falls back to `deepcopy` which both give a more complex output with e.g. `@code_native recursivecopy(3)` and `@code_native deepcopy(3)` than `@code_native copy(3)`? Should we just define `recursivecopy(x::Number) = copy(x)` in RecursiveArrayTools which by default just returns `x` (https://github.com/JuliaLang/julia/blob/master/base/number.jl#L88)?

ChrisRackauckas: Yes. There shouldn't be a need to `deepcopy` a number anymore. This could improve performance in some places as well.

ChrisRackauckas: We should make sure Measurements.jl still works after changing to `copy`. It should.

PR #96, opened 2019-01-20
Update nothing type checks

devmotion:
Fix checks that I missed in https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/91.


devmotion:
Ah it's again the same test errors - it succeeded in one out of four Travis runs (one of the test on nightly) but failed in the other ones. I have to check what's happening there in more detail.

PR #100, opened 2019-02-09
Add alias_u0 keyword

devmotion:
As discussed in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/655.


src/solve.jl
```diff
@@ -113,8 +113,16 @@ function DiffEqBase.__init(
 
     # create separate copies u and uprev, not pointing to integrator.u or integrator.uprev,
     # to cache uprev with correct dimensions and types
-    u = recursivecopy(integrator.u)
-    uprev = recursivecopy(integrator.uprev)
+    if typeof(prob.u0) <: Tuple
+        u = ArrayPartition(prob.u0,Val{true})
```
ChrisRackauckas: We should really be doing the conversion to ArrayPartitions inside of DiffEqBase instead, but that's a whole different story.

devmotion:
Tests fail due to https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/656.

Issue #97, opened 2019-01-20
Test failures of lazy interpolants


devmotion:
Currently some of the tests for Vern6 and Vern9 fail sometimes (see https://travis-ci.org/JuliaDiffEq/DelayDiffEq.jl/builds). However, the test failures do not depend on the Julia version (they both fail and succeed on Julia 1.0.3 and the nightlies from time to time).

Closer inspection reveals that tests only error when tested on `Intel(R) Xeon(R) CPU @ 2.30GHz` with `libLLVM-6.0.1 (ORCJIT, haswell)` but run without problems on `Intel(R) Xeon(R) CPU @ 2.50GHz` with `libLLVM-6.0.0 (ORCJIT, ivybridge)` and `Intel(R) Xeon(R) CPU @ 2.60GHz` with `libLLVM-6.0.0 (ORCJIT, sandybridge)`. So it seems most likely the problems are caused by some floating point differences in the in-place and not in-place implementations of Vern6 and Vern9 that become apparent on specific architectures and maybe are related to https://github.com/JuliaDiffEq/DelayDiffEq.jl/issues/73.

ChrisRackauckas:
That's pointing to a difference in compiler optimization, especially since muladd is used.

Issue #73, opened 2018-07-15
Approximate equality in not-in-place stepping


ChrisRackauckas:
This commit https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/72/commits/d3c5baeaf35e94433e3a5fd8075d0d3c15067f8b turned exact checks to approximate checks because the inplace vs outofplace versions now only step within floating point error. This is a minor point would could be corrected again in the future since on v0.6 it was completely error free.

ChrisRackauckas:
https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/72/commits/d0a7c60598a1634332978143c05c9e3c157ac01d

ChrisRackauckas:
https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/72/commits/8c8885d30f2a92875608252ea8451932bf681364

ChrisRackauckas:
Could potentially have been fixed by https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/commit/b86aae7eb0cddaa1e8fb76c75a67eac366e26ebb

devmotion:
I observed that (at least) one approximate check is required due to the fact that interpolating an ODE solution returns different values for no, scalar, and vector indices in some cases (on Julia 1.0.2, current master branch of DiffEqBase, OrdinaryDiffEq, and DelayDiffEq):
```julia
using DelayDiffEq, DiffEqProblemLibrary.DDEProblemLibrary
DDEProblemLibrary.importddeproblems()

sol = solve(prob_dde_1delay, MethodOfSteps(Tsit5()))

sol(7) # -0.05731721456328998
sol(7; idxs=1) # -0.05731721456328997
sol(7; idxs=[1])[1] # -0.05731721456328998
```
Hence (some of) the approximate checks might be caused by inconsistent interpolations in OrdinaryDiffEq, I guess. In that case it should be possible to observe the same approximate equalities when solving ODEs (which I haven't tried).

devmotion:
[...]

PR #101, opened 2019-02-09
Disable Vern6 and Vern9 tests on Haswell CPUs

devmotion:
This PR fixes https://github.com/JuliaDiffEq/DelayDiffEq.jl/issues/97 by disabling the two failing tests of Vern6 and Vern9 on Haswell CPUs only (compare previous builds https://travis-ci.org/JuliaDiffEq/DelayDiffEq.jl/jobs/482509932 with https://travis-ci.org/JuliaDiffEq/DelayDiffEq.jl/jobs/491089819).

Tests fail due to https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/656.

### DIFF
--- a/test/lazy_interpolants.jl
+++ b/test/lazy_interpolants.jl
@@ -15,7 +15,8 @@ include("common.jl")
 
             sol2 = solve(prob_notinplace, MethodOfSteps(Vern6()))
 
-            @test sol.t ≈ sol2.t && sol[1, :] ≈ sol2.u
+            # fails on Haswell CPUs: https://github.com/JuliaDiffEq/DelayDiffEq.jl/issues/97
+            Sys.CPU_NAME == "haswell" || @test sol.t ≈ sol2.t && sol[1, :] ≈ sol2.u
         end
 
         # Vern7
@@ -54,7 +55,8 @@ include("common.jl")
 
             sol2 = solve(prob_notinplace, MethodOfSteps(Vern9()))
 
-            @test sol.t ≈ sol2.t && sol[1, :] ≈ sol2.u
+            # fails on Haswell CPUs: https://github.com/JuliaDiffEq/DelayDiffEq.jl/issues/97
+            Sys.CPU_NAME == "haswell" || @test sol.t ≈ sol2.t && sol[1, :] ≈ sol2.u
         end
     end
 


ChrisRackauckas:
Is this just a too low tolerance thing? Why would Haswell matter?

devmotion:
Actually I don't know. It's difficult to debug this issue since I cannot reproduce the failures locally. As mentioned in https://github.com/JuliaDiffEq/DelayDiffEq.jl/issues/97, the errors are CPU specific and so I assume they are caused by floating point issues and muladd, similar to what I observed in https://github.com/JuliaDiffEq/DelayDiffEq.jl/issues/73.

I think it's helpful to get rid of these test errors since they could mask other errors such as in https://travis-ci.org/JuliaDiffEq/DelayDiffEq.jl/jobs/487827113.

ChrisRackauckas:
How far off is it on Haswell, small? Or large? If small, we should `if Haswell` and then just increase the tolerance a bit. That could be because of `@muladd` not applying on some Haswell chips. If it's large, that's worrisome.

devmotion:
I did some debugging here: https://travis-ci.org/JuliaDiffEq/DelayDiffEq.jl/jobs/491106802

For Vern6 the printed values of the following time points (out of 60 time points) are different:
```julia
julia> t1[t1 .!== t2]
6-element Array{Float64,1}:
 5.59595
 6.77066
 7.54998
 8.1625 
 9.09498
 9.86356

julia> t2[t1 .!== t2]
6-element Array{Float64,1}:
 5.59609
 6.76899
 7.55189
 8.16518
 9.10246
 9.87058
```
With the following `u` values:
```julia
julia> u1[u1 .!== u2]
6-element Array{Float64,1}:
  0.223364  
 -0.0157241 
 -0.109774  
 -0.0916953 
  0.00505008
  0.0513813 

julia> u2[u1 .!== u2]
6-element Array{Float64,1}:
  0.223349  
 -0.0153893 
 -0.109834  
 -0.0914809 
  0.00577223
  0.0515117 
```
Interestingly the printed values at 6.0, 7.0, and 10.0 are the same.

devmotion:
I get similar results for Vern9 (out of 27 time points):
```julia
julia> t1[t1 .!== t2]
2-element Array{Float64,1}:
 6.83966
 7.6147 

julia> t2[t1 .!== t2]
2-element Array{Float64,1}:
 6.83979
 7.55545

julia> u1[u1 .!== u2]
2-element Array{Float64,1}:
 -0.0292791
 -0.111386 

julia> u2[u1 .!== u2]
2-element Array{Float64,1}:
 -0.0293037
 -0.109977 
```

The printed values at time points 7.0, 8.0, 9.0, and 10.0 are the same.

By increasing the relative tolerances to 1e-3 for `t` and `u` with Vern6 and `u` with Vern9 and 1e-2 for `t` with Vern9 I can achieve approximate equality.

ChrisRackauckas:
Yeah seems like it could just be floating point accumulation when `@muladd` isn't done and the tolerances are high. Tests still fail though.

devmotion:
The test failures are caused by https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/656. The logs show that the "lazy interpolants" test pass on Haswell now.

devmotion:
I think a better fix might be to test for approximate equality of the solution at fixed time points, e.g., `t = 0:0.1:10`. I did some tests here https://travis-ci.org/JuliaDiffEq/DelayDiffEq.jl/jobs/491214135 and it seems that with Haswell CPUs `norm(first.(sol1(ts)) - sol2(ts))` is around `5.431765878617846e-9` and `3.1559293119007873e-13` for Vern6 and Vern9, respectively. Of course, a disadvantage of that approach would be that it's not clear anymore that both implementations are equivalent.

PR #102, opened 2019-02-12
Fix Vern6 and Vern9 test errors

devmotion:
That's a different approach to fix https://github.com/JuliaDiffEq/DelayDiffEq.jl/issues/97, as discussed in https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/101#issuecomment-462131423. In contrast to https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/101, the number of tests does not depend on the CPU. A disadvantage is that we do not test for (approximate) equality of the time steps in the in-place and out-of-place algorithms of Vern6 and Vern9 anymore.

### DIFF
--- a/test/lazy_interpolants.jl
+++ b/test/lazy_interpolants.jl
@@ -4,6 +4,7 @@ include("common.jl")
     @testset "simple problems" begin
         prob_inplace = prob_dde_1delay
         prob_notinplace = prob_dde_1delay_scalar_notinplace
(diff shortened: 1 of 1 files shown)


devmotion:
Tests for Vern6 and Vern9 pass on Haswell as the logs show. The test errors are due to https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/656.

PR #103, opened 2019-02-24
Reflect the DEStats changes

YingboMa:


src/solve.jl
```diff
@@ -343,15 +343,15 @@ function solve!(integrator::DDEIntegrator)
             typeof(prob),typeof(integrator.sol.alg),typeof(interp)}(
                 sol_array.u, u_analytic, errors, sol_array.t, interp.ks,
                 prob, integrator.sol.alg, interp, interp.alg_choice,
-                interp.dense, integrator.sol.tslocation, :Success)
+                interp.dense, integrator.sol.tslocation, integrator.integrator.sol.destats, :Success)
```
ChrisRackauckas: why from integrator.integrator?

YingboMa: `integrator.sol.destats` gives 0 for some reason.

ChrisRackauckas: isn't `integrator` being stepped, and `integrator.integrator` just being a store? @devmotion

devmotion: Yes, `integrator` is stepped but the solution in `integrator.integrator` is used for inter- and extrapolation and updated in https://github.com/JuliaDiffEq/DelayDiffEq.jl/blob/master/src/integrator_interface.jl#L25. Shouldn't you get `integrator.sol.destats == integrator.integrator.sol.destats` if you specify `destats` explicitly in https://github.com/JuliaDiffEq/DelayDiffEq.jl/blob/master/src/solve.jl#L72-L80?

ChrisRackauckas: Yes, that would do it. I agree that the correct thing to do would be to alias the destats here

ChrisRackauckas:
This passed locally, so I'm at least going to merge and tag since right now everything is broken. I think it just needed the nonlinear solver changes.

PR #105, opened 2019-06-01
Fixes `unwrap_cache` error

kanav99:
Fixes JuliaDiffEq/diffeqpy#49.
@tkf that should get fix once this is merged and tagged. Please report me If more such errors persist, as we have refactored nlsolve recently and there may be more hidden bugs.
Also @ChrisRackauckas we should move to the new Project.toml way for this repo too.


ChrisRackauckas:
Oh god, this found another bug which looks deeper.

ChrisRackauckas:
We need a way to trigger DelayDiffEq tests for OrdinaryDiffEq tags again.

kanav99:
This seems to be similar to the bugs we got in StochasticDiffEq after porting of nlsolve. The new nlsolve has some issues with floating point management which is not at all visible.

PR #108, opened 2019-06-02
Test Failure Fixes

kanav99:
Something weird happened between 21 March - 8 April in OrdinaryDiffEq due to which errors have increased but steps have decreased. DelayDiffEq was also hit by this. And its not algorithm specific, but for all the algorithms.

### DIFF
--- a/src/callbacks.jl
+++ b/src/callbacks.jl
@@ -128,7 +128,7 @@ function DiffEqBase.find_callback_time(integrator::DDEIntegrator, callback::Disc
         end
     end
 
-    tmin, order, contains_discontinuity
+    tmin, order, contains_discontinuity, 1
 end
 
 """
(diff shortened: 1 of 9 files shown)


ChrisRackauckas:
https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/commit/5263f5be5248cfff464ec80037c987df765fc493 made cases where `du = 0` have a more stable inital `dt`, but that would've changed the initial `dt` as well.

kanav99:
Yes this is it. This is the culprit. I tried running tests before and after the commit and it passes before that!

kanav99:
Commenting out those two lines made all the tests pass. I will understand the AD problem and make a patch there. We might need to tighten the tolerances in OrdinaryDiffEq tests too.

ChrisRackauckas:
The OrdinaryDiffEq.jl change should be harmless here.

kanav99:
On turning Off the AD fix, the tests pass, what should I do? Increase the estimates or make a fix in OrdinaryDIffEq? Or maybe a special implementation of `ode_determine_initdt`?

ChrisRackauckas:
Increase the estimates. It means they were too tight: they shouldn't fail due to a change in init dt.

kanav99:
I have marked a test broken because this PR was getting long and master is broken right now. I will handle the broken test in a seperate PR and merge this once this passes tests.

ChrisRackauckas:
Sounds good. Yeah seems like it's just a bunch of too strict regression tests

Issue #110, opened 2019-06-20
DDE callbacks are not initialized


alanderos91:
Please see https://github.com/JuliaDiffEq/DiffEqJump.jl/issues/71#issuecomment-504161383.

PR #111, opened 2019-06-20
Initialize DDEIntegrator callbacks

alanderos91:
Addresses #110. Issue detected with a DDE/Jump hybrid problem. Should tests live here or in DiffEqJump? I think the latter makes sense. Tests run locally, but I'm not sure if these changes break anything I'm not aware of...


ChrisRackauckas:
Thanks!

Issue #114, opened 2019-07-01
Change interface of analytical solutions


devmotion:
As discussed in https://github.com/JuliaDiffEq/DiffEqProblemLibrary.jl/pull/39:

The interface `f_analytic(u0, p, t)` doesn't make much sense, the analytical solution should depend on the history function as well. I propose using `f_analytic(u0, h, p, t)`, similar to the interface for `f(u0, h, p, t)`.

This change requires PRs to DiffEqBase, DiffEqProblemLibrary, and DelayDiffEq.

PR #115, opened 2019-07-01
Update tests and change interface of analytical solutions

devmotion:
Supersedes https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/113 and fixes https://github.com/JuliaDiffEq/DelayDiffEq.jl/issues/114.

Depends on https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/259 and https://github.com/JuliaDiffEq/DiffEqProblemLibrary.jl/pull/42.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DelayDiffEq"
 uuid = "bcd4f6db-9728-5f36-b5f7-82caef46ccdb"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>"]
-version = "v5.4.1"
+version = "5.4.1"
 
 [deps]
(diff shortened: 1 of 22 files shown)


devmotion:
Actually, my comment above is not correct: the changes in `solve.jl` move large parts of the construction of the final solution to DiffEqBase. That's the reason for why the tests do not fail even though both PRs in DiffEqBase and DiffEqProblemLibrary are not merged yet - the package still uses the old problem definitions and the old logic for computing the analytical solution in DiffEqBase.

So rather the other two PRs depend on each other.

devmotion:
I don't understand why but again test errors are showing up on Win32. Everything else passes.

However, we should probably try to rerun the tests with the updated versions of DiffEqBase and DiffEqProblemLibrary.

Issue #120, opened 2019-07-05
Slowdown from discontinuity declaration


ChrisRackauckas:
https://discourse.julialang.org/t/declaring-state-dependent-delay-with-ddeproblem-simulation-gets-slower/26048

It would be good to profile this a bit.

devmotion:
I had a quick look and posted some comments.

ChrisRackauckas:
Are we appropriately handling the type instability from the tuple of functions? I don't see why we would get so much extra allocations from the tracking, since that callback part should be non-allocating.

devmotion:
In general, we might want to use some of the ideas from the discussion of break point detection in RADAR5: http://www.unige.ch/~hairer/preprints/break-point.pdf

It seems two major differences are that
1) breaking point detection is only run if a step is rejected AND there exists a previous breaking point `T` such that `lag(u(t), p, t) - T` changes sign in the current step
2) estimation of the breaking points is done by iteratively updating the stage values of the current step and performing root finding based on the updated dense output.

ChrisRackauckas:
I don't know if we can do (2) with the current setup. It seems to really require implicit schemes. It would be nice to make a pure Julia RADAR5 and see how that performs though.

devmotion:
It seems applying suggestion (1) (and another one not mentioned in the comment above) is very helpful for this problem: https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/121

devmotion:
@AlexNascou Have you checked if your issues are fixed in the latest version of DelayDiffEq?

AlexNascou:
Yes I tried it last night and it worked perfectly after updating all the packages. And it was only marginally slower with the dependent lags declared.
Thanks again!

devmotion:
Great to hear, then I'll close this issue.

PR #121, opened 2019-07-09
Change tracking of discontinuities arising from dependent delays

devmotion:
This is a major change. Currently discontinuity tracking is performed by a special callback. However, this has several drawbacks:
- It is only run if the step is accepted.
- All possible combinations of lags and previous discontinuities have to be evaluated in order to determine the first discontinuity and to check whether the event of the discontinuity callback happens before any other continuous callback.
With this PR the behaviour is now:
- If a step is not accepted, then we check if we step over a discontinuity in the current integration step and try to estimate its time point, if existent.
- We do not check all possible combinations of previous discontinuities and dependent delays but just add the first one that is found to the list of tracked discontinuities.
Both approaches occur also in the algorithm of RADAR5, but there the time point is estimated in a different way.

Some examples:
- For the problem in https://github.com/JuliaDiffEq/DelayDiffEq.jl/issues/120 / https://discourse.julialang.org/t/declaring-state-dependent-delay-with-ddeproblem-simulation-gets-slower/26048, I obtain with the latest registered version of DelayDiffEq
```julia
julia> simulate(MethodOfSteps(Rosenbrock23()), 1.0, 3650.0; with_dependent_lags = false)
  0.188017 seconds (2.70 M allocations: 63.264 MiB, 7.62% gc time)

julia> simulate(MethodOfSteps(Rosenbrock23()), 1.0, 3650.0; with_dependent_lags = true)
┌ Warning: Interrupted. Larger maxiters is needed.
└ @ DiffEqBase ~/.julia/packages/DiffEqBase/P5x3s/src/integrator_interface.jl:150
┌ Warning: Interrupted. Larger maxiters is needed.
└ @ DiffEqBase ~/.julia/packages/DiffEqBase/P5x3s/src/integrator_interface.jl:150
 58.413165 seconds (946.67 M allocations: 18.846 GiB, 6.05% gc time)
```
for the simulation without and with dependent delays, whereas with this PR I get
```julia
julia> simulate(MethodOfSteps(Rosenbrock23()), 1.0, 3650.0; with_dependent_lags = false)
  0.183031 seconds (2.70 M allocations: 63.264 MiB, 7.33% gc time)

julia> simulate(MethodOfSteps(Rosenbrock23()), 1.0, 3650.0; with_dependent_lags = true)
  0.206473 seconds (3.11 M allocations: 69.686 MiB, 7.52% gc time)
```

- Simulating a state-dependent DDE by Waltman (see https://github.com/JuliaDiffEq/DiffEqProblemLibrary.jl/blob/master/src/dde/radar5.jl#L127-L163) with `Rosenbrock23` and `reltol = 1e-6` fails on the latest version of DelayDiffEq
```julia
using DelayDiffEq
using DiffEqProblemLibrary.DDEProblemLibrary

DDEProblemLibrary.importddeproblems()

julia> @time sol = solve(DDEProblemLibrary.prob_dde_RADAR5_waltman, MethodOfSteps(Rosenbrock23()); reltol = 1e-6);
┌ Warning: Interrupted. Larger maxiters is needed.
└ @ DiffEqBase ~/.julia/packages/DiffEqBase/P5x3s/src/integrator_interface.jl:150
 46.966992 seconds (660.92 M allocations: 13.772 GiB, 6.29% gc time)

julia> sol.t[end]
119.0

julia> sol.u[end]
6-element Array{Float64,1}:
   2.781833021386451e-17 
   5.972840495643085e-5  
   7.499999999997218e-5  
   1.0658141036401504e-27
 118.99916746413781      
   1.9126931099472524e-6 

julia> length(sol.t)
679
```

image

image

whereas with this PR I get
```julia
julia> @time sol = solve(DDEProblemLibrary.prob_dde_RADAR5_waltman, MethodOfSteps(Rosenbrock23()); reltol = 1e-6);
  0.120684 seconds (1.68 M allocations: 33.297 MiB, 3.81% gc time)

julia> sol.t[end]
300.0

julia> sol.u[end]
6-element Array{Float64,1}:
  -2.6474114243335877e-44
   5.9716314262430934e-5 
   7.500000000000051e-5  
   0.0005210301012929696 
 299.9990613755263       
 269.3175356798877       

julia> length(sol.t)
1083
```

image

image

As a comparison, running Guglielmi's and Hairer's Fortran code (after adjusting the parameters) with RADAR5 (with slightly different tolerances - they set the absolute tolerances of the first three components to 1e-18 when an element-wise relative tolerance of 1e-6 is specified) yields
```
 ELAPSED TIME=  0.246417999

 X =  300.000000    Y =  0.4164223440E-50  0.5999999615E-04  0.7500000000E-04  0.5227343056E-03  0.3000000000E+03  0.2807189083E+03
```


devmotion:
CI tests fail since https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/822 is a requirement. Moreover, the integrator tests fail currently due to (probably) some issues in OrdinaryDiffEq - they fail even on master although some days ago they could be run successfully.

ChrisRackauckas:
This is a good idea with a shockingly clean implementation.

ChrisRackauckas:
it seems like the issues are just the known SDIRK Newton test failures.

PR #122, opened 2019-07-10
WIP: Fix for nlsolver changes

ChrisRackauckas:
Test case:

```julia
using DelayDiffEq, Test
using DiffEqProblemLibrary.DDEProblemLibrary

DDEProblemLibrary.importddeproblems()

using DiffEqProblemLibrary.DDEProblemLibrary:
  # DDE problems with 1 constant delay
  prob_dde_constant_1delay_ip, prob_dde_constant_1delay_oop, prob_dde_constant_1delay_scalar,
  prob_dde_constant_1delay_long_ip, prob_dde_constant_1delay_long_oop, prob_dde_constant_1delay_long_scalar,
  # DDE problems with 2 constant delays
  prob_dde_constant_2delays_ip, prob_dde_constant_2delays_oop, prob_dde_constant_2delays_scalar,
  prob_dde_constant_2delays_long_ip, prob_dde_constant_2delays_long_oop, prob_dde_constant_2delays_long_scalar,
  # Mackey-Glass equation (model of blood production) with 1 constant delay
  prob_dde_DDETST_A1

const prob_ip = prob_dde_constant_1delay_ip
const prob_scalar = prob_dde_constant_1delay_scalar
const ts = 0:0.1:10

alg = ImplicitEuler()
println(nameof(typeof(alg)))

stepsalg = MethodOfSteps(alg)
sol_ip = solve(prob_ip, stepsalg)
sol_scalar = solve(prob_scalar, stepsalg, reltol=1e-6)

@test sol_ip(ts, idxs=1) ≈ sol_scalar(ts)
@test sol_ip.t ≈ sol_scalar.t && sol_ip[1, :] ≈ sol_scalar.u

using Plots
plot(sol_ip)
plot!(sol_scalar)
```

The big change that occurred is that now instead of using a pure macro-based creation of the nonlinear solvers https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/master/src/nlsolve/utils.jl#L98 , we now have proper constructors: https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/master/src/nlsolve/utils.jl#L237 . So we just need to make these two things not alias. I almost have it, but for some reason it's only slightly off now.

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -169,7 +169,7 @@ function DiffEqBase.__init(prob::DiffEqBase.AbstractDDEProblem,
         dde_cache = OrdinaryDiffEq.CompositeCache(caches, alg.alg.choice_function, 1)
     else
         dde_cache = build_linked_cache(integrator.cache, alg.alg, u, uprev, uprev2, dde_f,
-                                       tspan[1], dt, p)
+                                       tspan[1], dt, p, uEltypeNoUnits)
     end
 
     # filter provided discontinuities
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -164,7 +164,7 @@ function DiffEqBase.__init(prob::DiffEqBase.AbstractDDEProblem,
     # new cache with updated u, uprev, uprev2, and function f
     if typeof(alg.alg) <: OrdinaryDiffEq.OrdinaryDiffEqCompositeAlgorithm
         caches = map((x,y) -> build_linked_cache(x, y, u, uprev, uprev2, dde_f,
-                                                 tspan[1], dt, p),
+                                                 tspan[1], dt, p, uEltypeNoUnits),
                      integrator.cache.caches, alg.alg.algs)
         dde_cache = OrdinaryDiffEq.CompositeCache(caches, alg.alg.choice_function, 1)
     else
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -70,6 +70,13 @@ assign_expr(::Val{name}, ::Type{<:DiffEqDiffTools.GradientCache},
           ::Type) where {name} =
               :($name = OrdinaryDiffEq.build_grad_config(alg, f, tf, du1, t))
 
+# nlsolver
+assign_expr(::Val{name}, ::Type{<:OrdinaryDiffEq.NLSolver{true}}, ::Type) where name =
+    :($name = DiffEqBase.iipnlsolve(alg,u,uprev,p,t,dt,f,W,
+                                    cache.$name.du1,uEltypeNoUnits,
+                                    uEltypeNoUnits,cache.$name.γ,cache.$name.c))
+    # Second uEltypeNoUnits should be uBottomEltypeNoUnits
+
 # update implicit RHS
 assign_expr(::Val{name}, ::Type{<:OrdinaryDiffEq.ImplicitRHS}, ::Type) where name =
     :($name = OrdinaryDiffEq.ImplicitRHS(f, cache.tmp, t, t, t, cache.dual_cache,p))
@@ -95,10 +102,9 @@ assign_expr(::Val{name}, ::Type{<:NLSolversBase.OnceDifferentiable},
 Create cache for algorithm `alg` from existing cache `cache` with updated `u`, `uprev`,
 `uprev2`, `f`, `t`, and `dt`.
 """
-@generated function build_linked_cache(cache, alg, u, uprev, uprev2, f, t, dt,p)
+@generated function build_linked_cache(cache, alg, u, uprev, uprev2, f, t, dt, p, uEltypeNoUnits)
     assignments = [assign_expr(Val{name}(), fieldtype(cache, name), cache)
                    for name in fieldnames(cache) if name ∉ [:u, :uprev, :uprev2, :t, :dt]]
-
     :($(assignments...); $(DiffEqBase.parameterless_type(cache))($(fieldnames(cache)...)))
 end
 


ChrisRackauckas:
@kanav99 now that https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/caches/sdirk_caches.jl#L15-L19 are all in the nlsolver, do we need them directly in the cache?

kanav99:
Nope, we actually dont need it, but we have to remove them carefully from the alg cache. These fields are interchangeably used from cache or nlcache at random.

devmotion:
IMO we should definitely get rid of this redundancy, it makes it not only a lot more difficult in DelayDiffEq to ensure that objects are aliased correctly but also more error prone and confusing in OrdinaryDiffEq.

ChrisRackauckas:
I agree. Is there a better solution? I am not sure if we can just deepcopy the cache?

devmotion:
The problem with here is that in the construction of the NLSolver the function `f` is passed around. Hence we have to repeat all calls and should do something like
```julia
# nlsolver
# @getiipnlsolvefields contains the line fsalfirst = zero(rate_prototype)
# which fails since rate_prototype is not available
assign_expr(::Val{name}, ::Type{<:OrdinaryDiffEq.NLSolver{true}}, ::Type) where name =
  quote
    J, W = OrdinaryDiffEq.iip_generate_W(alg,u,uprev,p,t,dt,f,recursive_unitless_eltype(u))
    $name = DiffEqBase.iipnlsolve(alg,u,uprev,p,t,dt,f,W,cache.$name.du1,
                                  recursive_unitless_eltype(u),
                                  recursive_unitless_bottom_eltype(u),cache.$name.γ,
                                  cache.$name.c)
    @unpack z,dz,tmp,k,uf,du1,jac_config,linsolve = nlsolver
    b = nlsolver.ztmp
end
assign_expr(::Val{name}, ::Type{<:OrdinaryDiffEq.NLSolver{false}}, ::Type) where name =
  quote
    W = OrdinaryDiffEq.oop_generate_W(alg,u,uprev,p,t,dt,f,recursive_unitless_eltype(u))
    $name = DiffEqBase.oopnlsolve(alg,u,uprev,p,t,dt,f,W,cache.$name.du1,
                                  recursive_unitless_eltype(u),
                                  recursive_unitless_bottom_eltype(u),cache.$name.γ,
                                  cache.$name.c)
    DiffEqBase.@getoopnlsolvefields
end
```
I had hoped that would be sufficient since `nlsolver` seems to be the last field of all (?) caches, and hence all other incorrectly generated assignments or constructions of other fields should not matter in the end, I assumed. Unfortunately that alone does not seem to fix the issue.

devmotion:
Right now I try to do it in a completely different way by constructing a dummy integrator (instead of a full ODE integrator) and doing the proper cache construction only with the DDE integrator. But I'm not sure if I can work around all dependencies and if it's actually possible to get rid of the cache construction/copy.

devmotion:
I think I spent too much time trying to fix this in DelayDiffEq - I'm convinced there's an issue in OrdinaryDiffEq. Both on master and with the latest released versions of OrdinaryDiffEq and DiffEqBase the following errors
```julia
using OrdinaryDiffEq, Test

function f_ip(du, u, p, t)
    du[1] = - u[1]
    nothing
end

f_scalar(u, p, t) = -u


const prob_ip = ODEProblem(f_ip, [1.0], (0.0, 10.0))
const prob_scalar = ODEProblem(f_scalar, 1.0, (0.0, 10.0))
const ts = 0:0.1:10

alg = ImplicitEuler()
println(nameof(typeof(alg)))

sol_ip = solve(prob_ip, alg)
sol_scalar = solve(prob_scalar, alg)

@test sol_ip(ts, idxs=1) ≈ sol_scalar(ts)
@test sol_ip.t ≈ sol_scalar.t && sol_ip[1, :] ≈ sol_scalar.u
```
and yields the error message
```julia
Test Failed
  Expression: sol_ip(ts, idxs=1) ≈ sol_scalar(ts)
   Evaluated: t: 0.0:0.1:10.0
u: [1.0, 0.702711, 0.566709, 0.384261, 0.348434, 0.2894, 0.220095, 0.182301, 0.165377, 0.126175  …  5.11408e-7, 4.9757e-7, 4.85718e-7, 4.74637e-7, 4.63111e-7, 4.49927e-7, 4.33869e-7, 4.13723e-7, 3.88273e-7, 3.56306e-7] ≈ t: 0.0:0.1:10.0
u: [1.0, 0.905837, 0.820647, 0.743467, 0.673701, 0.610487, 0.55313, 0.50118, 0.454112, 0.411444  …  0.000130758, 0.000118791, 0.000107958, 9.81271e-5, 8.91296e-5, 8.10186e-5, 7.36185e-5, 6.7053e-5, 6.10239e-5, 5.56887e-5]
ERROR: There was an error during testing
```

devmotion:
I just tried to update the cache creation such that it works with the latest NLSolver updates in OrdinaryDiffEq such as https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/caches/sdirk_caches.jl#L15-L16, similar to what I outlined in https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/122#issuecomment-510011389. However, it seems it's almost impossible to do this a reasonable way since it is completely unpredictable which fields of `nlsolver` should be aliased with some other fields in `cache`, such as https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/caches/sdirk_caches.jl#L173.

IMO we should get rid of these cache hacks completely. It seems unreasonable to create a cache with an incorrect function that is never used and try to base a cache which is used and has to be guaranteed to be correct on this incorrect cache. Currently, my impression is that we only need the type and the tableaus of the dummy caches for dispatching and computing the interpolation. Thus I propose:
1. Define a method `tableau(alg, uBottomEltypeNoUnits, tTypeNoUnits)` (I just checked `sdirk_caches`, maybe other arguments needed?) in OrdinaryDiffEq for every algorithm which just returns `nothing` by default.
2. Automatically forward the currently existing methods `alg_cache` to `alg_cache` with an additional tableau argument, so it's possible to reuse the tableau.
3. Add to the `@cache` macro a method `alg_cache(alg, tableau, Val{true/false})` that sets all other fields except the one of name `tab` to `nothing`.
Then we would get a completely reduced cache for the dummy integrator with at most the tableau that we can even reuse for the actually used integrator. Moreover, the full integrator would be created in exactly the same way as in OrdinaryDiffEq.

ChrisRackauckas:
Goodbye fake caches lol

PR #124, opened 2019-07-12
Massive change

devmotion:
This PR introduces some major changes in the construction of the integrator and its solution. In my opinion these modifications clarify the algorithm a lot since we get rid both of an unneeded layer of functions and interpolations in the initialization of the integrator and the complete DelayDiffEq-specific logic for building the solution at the end of the integration.

Instead of constructing the solution in the end from a single dense solution that is modified throughout the integration, we now clearly separate the solution and the dense history at every point of the algorithm and build the solution over time, according to the parameters specified by the user. Apart from the increased readability, this has several other advantages:
- It leads to a handling of solutions that is consistent with all other DiffEq packages (e.g., the handcrafted implementation in DelayDiffEq deals with `save_end`, `save_everystep`, etc. keywords in a slightly different way than OrdinaryDiffEq)
- We will get the iterator interface for free, which might be very helpful for step by step debugging
- Speculating a bit, I guess this makes it very easy to allow the user to specify "history indices", so that we do not need to keep the full dense history anymore (my guess is that this is only a matter of forwarding these indices to the ODE integrator and adding a `history_idxs` field to `HistoryFunction` such that we can handle calls of the history function in a correct way)

The only user-facing change so far is the removal of the `minimal_solution` keyword. Depending on whether we want to keep this functionality or not, I will either try to update the old implementation or add a warning for the users. Although interesting, this functionality always felt a bit intransparent to me.


devmotion:
Tests fails since this PR requires https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/829 (and since the SDIRK tests are still broken).

I tried to split parts of the algorithm into smaller functions, I guess, some of them could be reused (in a modified form) also in OrdinaryDiffEq.

ChrisRackauckas:
I took a quick look at it and I think it looks great. I put some fixes into DiffEqBase and OrdinaryDiffEq that might handle the SDIRK case.

ChrisRackauckas:
I'll do a more thorough review tomorrow morning.

devmotion:
The SDIRK tests are still broken :disappointed: The discontinuity tests at https://ci.appveyor.com/project/ChrisRackauckas/delaydiffeq-jl/builds/25981993/job/498lpcrt3hgbcecq took a ridiculously long time to run, and also on Travis there's a noticeable increase (395 sec (https://travis-ci.org/JuliaDiffEq/DelayDiffEq.jl/jobs/558865652) versus 226 sec before (https://travis-ci.org/JuliaDiffEq/DelayDiffEq.jl/jobs/557136763)) - could this be due to the method warnings?

ChrisRackauckas:
I don't think the method warnings would cause that.

devmotion:
Hmm, it seems timings are increased not for the discontinuity tests specifically but for the first tests we run. Also the "constrained time steps" tests take longer (472 sec vs 305 sec before, https://travis-ci.org/JuliaDiffEq/DelayDiffEq.jl/jobs/558865653 vs https://travis-ci.org/JuliaDiffEq/DelayDiffEq.jl/jobs/557136764), but the other tests seem unaffected.

devmotion:
Should we wait with merging until we figure out what causes these increased times? Otherwise, apart from the SDIRK tests everything passes.

ChrisRackauckas:
Do the benchmarks show a regression?

devmotion:
I'll check that. The problem with the benchmarks currently is that due to the quite long compile time the first run takes a lot of time (that's not a problem with this PR specifically, the same can be observed with the latest registered version as well, see e.g. https://github.com/JuliaDiffEq/DiffEqDevTools.jl/issues/42 - and I had the (maybe incorrect) impression that compile time was a bit lower when I used this PR). I want to check whether fixing OrdinaryDiffEq at https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/commit/331f7742e1872516fa0f9fa81bf4eccef60b72dc (before SparseDiffTools was added) helps to reduce compile times.

devmotion:
I ran the Mackey and Glass benchmark with the master branch 
[Mackey_Glass_wpd.pdf](https://github.com/JuliaDiffEq/DelayDiffEq.jl/files/3398274/Mackey_Glass_wpd.pdf) and this PR [Mackey_Glass_wpd.pdf](https://github.com/JuliaDiffEq/DelayDiffEq.jl/files/3398270/Mackey_Glass_wpd.pdf). I cannot observe any regressions, on the contrary, usually times decreased with this PR (but maybe that's just spurious fluctuations). Moreover, the results look similar to http://juliadiffeq.org/DiffEqBenchmarks.jl/html/NonStiffDDE/Mackey_Glass_wpd.html.

However, it's really annoying to run these benchmarks at the moment since for each algorithm the initial run takes multiple minutes, whereas subsequent runs are a matter of seconds. E.g., computing  the first set of low order RK methods with high tolerances and final error estimate took more than 15 minutes, and in total it took about half an hour to run all tests. Since my computer becomes almost unusable during these compilation phases I started to run these benchmarks on a server in our institute. There really should be a way to speed up the compilation.

ChrisRackauckas:
We have a server we benchmark on. I can see if we can get you access if you want. But yeah, I'm not sure why the benchmarks take so long to compile. I am not sure how to profile compilation. This might be a good Hackathon thing to corner Jeff with and find out.

PR #126, opened 2019-07-17
Implement iteration and resizing (and fix integration with fixed time steps)

devmotion:
This PR implements the iteration interface and enables resizing of the integrator. Additionally, it fixes a bug with fixed time steps (even with `adaptive = false` integration was performed with adaptive time steps) that I discovered by testing the iteration interface.


devmotion:
Seems the tests on AppVeyor take too long to run due to the Zygote issue. Interesting, that on 32bit we get a dtmin warning but not on all other systems - floating point issues again?

In general, we should update the tests if more resizing tests are added to OrdinaryDiffEq - I guess, we might miss some use cases/bugs at the moment.

ChrisRackauckas:
We are definitely missing cases in OrdinaryDiffEq at the moment. @kanav99 was looking at it.

PR #130, opened 2019-07-17
Fix the closure issue with the ODE function

YingboMa:


src/solve.jl
```diff
@@ -78,12 +78,10 @@ function DiffEqBase.__init(prob::DiffEqBase.AbstractDDEProblem,
   # we wrap the user-provided history function such that function calls during the setup
   # of the integrator do not fail
   if isinplace(prob)
-    ode_prob = ODEProblem{true}(ODEFunction{true}((du, u, p, t) -> f.f(du, u, h, p, t);
-                                                  mass_matrix = f.mass_matrix),
```
devmotion: We should enclose `h` here as well, otherwise function calls (e.g. when setting up a nonlinear solver) will fail.

devmotion:
I just found another closure that probably should be fixed as well (https://github.com/JuliaDiffEq/DelayDiffEq.jl/blob/f98741f11adfc86813ecdf76ac53d4ba5deac664/src/track.jl#L131-L133)

devmotion:
I was working on another fix in this branch `closure`, but I guess my ideas of introducing a callable struct `ODEFunctionWrapper` was not helpful since I couldn't fix the long compilation times on Appveyor https://ci.appveyor.com/project/devmotion/delaydiffeq-jl-ungfr/build/job/wunteqaioqb72kyg.

ChrisRackauckas:
I think the difficulty with the ODEFunctionWrapper approach is that I see that as another difficult thing to keep in sync with DiffEqBase, and will need a manual update every time anything else changes.

devmotion:
Keeping everything bundled together in a clean `struct` might be convenient if we would make use of Jacobians etc., but doesn't matter right now I guess.

I'm a bit worried, however, that this might not be the reason for the compilation issues. At least so far the times on Appveyor and Travis do not show any improvement.

src/solve.jl
```diff
@@ -77,11 +85,12 @@ function DiffEqBase.__init(prob::DiffEqBase.AbstractDDEProblem,
   #   current integration step (so the interpolation is fixed while updating the stages)
   # we wrap the user-provided history function such that function calls during the setup
   # of the integrator do not fail
+  _fh = HistoryWrapper(f, h)
```
devmotion: Should we just wrap `f.f`? To support Jacobians etc properly, we would have to add them to the surround ODEFunction, so we could just get rid of them here?

ChrisRackauckas: yeah, maybe we do need to wrap the whole thing.

PR #134, opened 2019-07-26
Use BinaryMinHeap

devmotion:
See https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/841.


devmotion:
Tests fail since the `resize!` changes in OrdinaryDiffEq are not released yet.

Issue #133, opened 2019-07-26
forward differentiation using `optimize` for DDE 


Farnazmdi:
Hello, 

I am working on an optimization problem for delay differential equation, and I wanted to use forward differentiation using `optimize(...; autodif:=forward)`, but it doesn't work. 
This is the error we get:
`
MethodError: Cannot `convert` an object of type Discontinuity{ForwardDiff.Dual{ForwardDiff.Tag{getfield(Main, Symbol("#residuals#13")){Int64,Array{Float64,2},Array{Float64,2},Array{Float64,1},Array{Float64,1}},Float64},Float64,6}} to an object of type Discontinuity{Float64}
Closest candidates are:
  convert(::Type{S}, !Matched::T<:(Union{CategoricalString{R}, CategoricalValue{T,R} where T} where R)) where {S, T<:(Union{CategoricalString{R}, CategoricalValue{T,R} where T} where R)} at /home/asm/.julia/packages/CategoricalArrays/xjesC/src/value.jl:91
  convert(::Type{T}, !Matched::T) where T at essentials.jl:154
  Discontinuity{Float64}(::Any, !Matched::Any) where tType at /home/asm/.julia/packages/DelayDiffEq/7Ng3E/src/discontinuity_type.jl:8

Stacktrace:
 [1] fill!(::SubArray{Discontinuity{Float64},1,Array{Discontinuity{Float64},1},Tuple{UnitRange{Int64}},true}, ::Discontinuity{ForwardDiff.Dual{ForwardDiff.Tag{getfield(Main, Symbol("#residuals#13")){Int64,Array{Float64,2},Array{Float64,2},Array{Float64,1},Array{Float64,1}},Float64},Float64,6}}) at ./multidimensional.jl:837
 [2] __cat(::Array{Discontinuity{Float64},1}, ::Tuple{Int64}, ::Tuple{Bool}, ::Array{Discontinuity{Float64},1}, ::Vararg{Any,N} where N) at ./abstractarray.jl:1412
 [3] _cat_t(::Val{1}, ::Type, ::Array{Discontinuity{Float64},1}, ::Vararg{Any,N} where N) at ./abstractarray.jl:1392
 [4] #cat_t#103(::Val{1}, ::Function, ::Type{Discontinuity{Float64}}, ::Array{Discontinuity{Float64},1}, ::Vararg{Any,N} where N) at ./abstractarray.jl:1384
[...]

PR #139, opened 2019-07-29
Add differentiation tests

devmotion:
I added tests similar to the ones in https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/test/interface/ad_tests.jl. Unfortunately they show that AD doesn't work (as suspected in https://github.com/JuliaDiffEq/DelayDiffEq.jl/issues/133).

### DIFF
--- a/test/differentiation.jl
+++ b/test/differentiation.jl
@@ -13,14 +13,14 @@ h(p, t) = [p[4]]
   function test(p)
     prob = DDEProblem(f, [p[5]], h, eltype(p).((0.0, 20.0)), copy(p);
                       constant_lags = [p[2]])
-    sol = solve(prob, MethodOfSteps(Tsit5()))
+    sol = solve(prob, MethodOfSteps(Tsit5()); abstol = 1e-14, reltol = 1e-14)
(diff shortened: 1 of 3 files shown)


test/differentiation.jl
```diff
@@ -0,0 +1,56 @@
...
+  nothing
+end
+
+h(p, t) = [p[4]]
+
+@testset "Gradient" begin
+  function test(p)
+    prob = DDEProblem(f, [p[5]], h, eltype(p).((0.0, 20.0)), copy(p);
+                      constant_lags = [p[2]])
+    sol = solve(prob, MethodOfSteps(Tsit5()))
```
ChrisRackauckas: you have to set the tolerance really low to test the difference here, since finite difference is really sensitive.

devmotion: Ah OK. However, I guess it does not matter at the moment - the values are off by multiple orders of magnitude, and even the signs do not match.

ChrisRackauckas: yeah that's bad haha

devmotion:
Tests on AppVeyor took too long to run, but the differentiation tests passed successfully.

PR #129, opened 2019-07-17
Help the compiler

devmotion:
The compiler seems to have a hard time figuring out the return type of function calls such as
```julia
init(DDEProblemLibrary.prob_dde_constant_1delay_ip, MethodOfSteps(Tsit5()))
```
even though we try to specify the type parameters of the DDEIntegrator. On the master branch, e.g., I get
```julia
julia> @code_warntype init(DDEProblemLibrary.prob_dde_constant_1delay_ip, MethodOfSteps(Tsit5()));
....
│    %51 = invoke %47(%22::DDEProblem{Array{Float64,1},Tuple{Float64,Float64},Array{Int64,1},Tuple{},true,DataType,DDEFunction{true,typeof(DiffEqProblemLibrary.DDEProblemLibrary.f_dde_constant_1delay_ip!),LinearAlgebra.UniformScaling{Bool},typeof(DiffEqProblemLibrary.DDEProblemLibrary.fanalytic_dde_constant_1delay),Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing,Nothing},typeof(DiffEqProblemLibrary.DDEProblemLibrary.h_dde_constant_ip),Nothing}, %1::MethodOfSteps{Tsit5,Nothing,Nothing,Nothing,false}, %48::Array{Array{Float64,1},1}, %49::Array{Float64,1}, %50::Array{Any,1})::DelayDiffEq.DDEIntegrator{Tsit5,true,_1,_2,_3,_4,_5,_6,_7,_8,_9,Array{Array{Float64,1},1},_10,_11,_12,_13,_14,_15,Float64,Int64,Array{Float64,1},_16,Nothing} where _16 where _15 where _14 where _13 where _12 where _11 where _10 where _9 where _8 where _7 where _6 where _5 where _4 where _3 where _2 where _1
└───       goto #28
28 ─       return %51
```
By just changing `p` to `prob.p` in https://github.com/JuliaDiffEq/DelayDiffEq.jl/blob/master/src/solve.jl#L258, the compiler is able to figure out that type parameter correctly. So I suspect that type inference breaks down at some point. I still don't understand why this becomes only apparent after having simplified our nested type structures a bit.

In this PR I tried reduce the number of non-inferred type parameters as much as possible and at the same time I replaced the full ODE integrator with a very reduced integrator that only contains a very reduced set of objects (and hence types) compared with a full ODE integrator. Additionally I started breaking the function into smaller pieces, hoping that this would help the compiler as well.

[...]

PR #140, opened 2019-07-30
Add HistoryODEIntegrator

devmotion:
This PR implements a reduced and cleaned version of https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/129 (some other features in that PR were implemented in the last days by other PRs).

The main idea is that by not using a full ODE integrator for the dummy integrator we can save a lot of allocations and type parameters. Initial tests show a huge decrease in test time and allocations compared with the master branch. However, I haven't tried to get rid of all type instabilities in `init`, so potentially further improvements are possible.

Some of the helper functions could/should be moved to OrdinaryDiffEq or DiffEqBase at some point, I guess.


devmotion:
On Travis the total time it takes to run the tests decreased by more than 50%: [from around 44 minutes (Julia 1.0) and 33 minutes (Julia 1.3)](https://travis-ci.org/JuliaDiffEq/DelayDiffEq.jl/builds/564905204) to [around 22 minutes (Julia 1.0) and 13 minutes (Julia 1.3)](https://travis-ci.org/JuliaDiffEq/DelayDiffEq.jl/builds/565455026?utm_source=github_status&utm_medium=notification).

devmotion:
The discrepancy between Julia 1.0 and Julia 1.3 can maybe explained by type instabilities.

Running
```julia
julia> @code_warntype init(DDEProblemLibrary.prob_dde_constant_1delay_ip, MethodOfSteps(Tsit5()));
```
shows that there are no type instabilities on Julia 1.3 whereas there are 6 type instabilities on Julia 1.1.1. I guess, some (maybe all?) of them can be fixed in similar ways as discussed in https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/129#issue-298605657. The problem of inferring the integrator options is gone, but the types of `p`, `resid`, `tdir`, `sol`, `f`, and `cache` cannot be inferred. By some tricks it was possible to infer at least the types of `p`, `resid`, and `tdir` in https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/129, but I haven't added these in this PR to keep it as clean as possible. I guess, those hacks could be added in a separate PR.

src/utils.jl
```diff
@@ -190,110 +255,97 @@ function Base.sizehint!(sol::DESolution, alg, tspan, tstops, saveat;
...
+  # obtain cache (we alias uprev2 and uprev)
+  ode_cache = OrdinaryDiffEq.alg_cache(alg.alg, ode_u, rate_prototype, uEltypeNoUnits,
+                                       uBottomEltypeNoUnits, tTypeNoUnits, ode_uprev,
+                                       ode_uprev, ode_f, t, zero(dt), reltol, p, calck,
+                                       Val{isinplace(prob)})
+
+  # build dense interpolation of history
+  if iscomposite(alg)
+    ode_alg_choice = Int[]
+    ode_id = OrdinaryDiffEq.CompositeInterpolationData(ode_f, ode_timeseries, ode_ts, ode_ks,
```
ChrisRackauckas: we might only need these interpolations

ChrisRackauckas:
having to do all of the options handling worries me, and I think we might only need the interpolations to do this correctly? The history function only needs to know how to get values, so if we do that and update it as necessary, it seems like we'd be good and we could drop the second integrator?

devmotion:
> having to do all of the options handling worries me

Which options do you mean?

> The history function only needs to know how to get values, so if we do that and update it as necessary, it seems like we'd be good and we could drop the second integrator?

The whole purpose of the second integrator is to be able to extrapolate and to interpolate in the current integration interval. Currently the cache is also used for dispatching on which interpolation is actually used and to get the tableaus necessary for interpolation. So I don't see how we could get rid of this? Or is there a shortcut possible somewhere in OrdinaryDiffEq? But even in that case we would need the tableau?

ChrisRackauckas:
>Which options do you mean?

All of the keyword arguments.

>The whole purpose of the second integrator is to be able to extrapolate and to interpolate in the current integration interval. Currently the cache is also used for dispatching on which interpolation is actually used and to get the tableaus necessary for interpolation. So I don't see how we could get rid of this? Or is there a shortcut possible somewhere in OrdinaryDiffEq? But even in that case we would need the tableau?

Currently the cache is used to dispatch because that's our dispatching system, but if we made the interpolants actually be different types then we could probably clean a bunch of stuff up, and as a side effect the dispatching for them would be independent of the integrator. Other than the `add_steps!` portions it would run completely automatically.

devmotion:
> Currently the cache is used to dispatch because that's our dispatching system, but if we made the interpolants actually be different types then we could probably clean a bunch of stuff up, and as a side effect the dispatching for them would be independent of the integrator. Other than the add_steps! portions it would run completely automatically.

Yes, I think that could work.

> All of the keyword arguments.

My ideal would be if the DDEIntegrator is a struct that just wraps an ODEIntegrator, a history, and a fixed point solver. Then hopefully most arguments could just be forwarded.

PR #141, opened 2019-07-30
Fix resize

devmotion:
When I was using the OrdinaryDiffEq master branch some days ago, the `resize!` tests were broken. This update fixed my issues back then.


devmotion:
Seems the new OrdinaryDiffEq breaks even more stuff in DelayDiffEq.

devmotion:
Because it requires less changes, I just confirmed with the branch https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/140 (without any changes, the Jacobian and resize tests there are also broken) that replacing the lines
```julia
if iscomposite(alg)
  caches = map((x, y) -> build_linked_cache(x, y, u, uprev, uprev2, f_with_history,
                                              t0, dt, p),
                 ode_cache.caches, alg.alg.algs)
    cache = OrdinaryDiffEq.CompositeCache(caches, alg.alg.choice_function, 1)
  else
    cache = build_linked_cache(ode_cache, alg.alg, u, uprev, uprev2,
                               f_with_history, t0, dt, p)
  end
```
with
```julia
  cache = OrdinaryDiffEq.alg_cache(alg.alg, u, rate_prototype, uEltypeNoUnits,
                                   uBottomEltypeNoUnits, tTypeNoUnits, uprev,
                                   uprev2, f_with_history, t0, zero(dt), reltol_internal, p, calck,
                                   Val{isinplace(prob)})
```
fixes the Jacobian test failures (the resize tests still require the changes in this PR). Thus the problem is caused by the cache changes in OrdinaryDiffEq in combination with the completely broken cache setup in DelayDiffEq (see https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/122).

As a quick fix, I suggest to just create a separate unlinked cache, as explained above. In the long run, we could try to reduce the allocations and unneeded setup of the dummy integrator by doing something along the lines of https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/122#issuecomment-515971409.

devmotion:
Closed in favor of https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/140.

PR #142, opened 2019-08-01
Add fixed-point solver

devmotion:
This PR introduces a dedicated fixed-point solver. Currently, the stopping and convergence criterion of the fixed-point iterations is a very ad-hoc heuristic. In this PR it is updated to basically the same logic as the non-linear solvers in DiffEqBase.

To reduce the amount of changes, intentionally this PR does not include Anderson acceleration. However, it should be straightforward to add it afterwards by more or less copying from the implementation of the non-linear solvers.

### DIFF
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -1,41 +1,21 @@
 abstract type DelayDiffEqAlgorithm <: AbstractDDEAlgorithm end
 abstract type AbstractMethodOfStepsAlgorithm{constrained} <: DelayDiffEqAlgorithm end
 
-struct MethodOfSteps{algType,AType,RType,NType,constrained} <:
-    AbstractMethodOfStepsAlgorithm{constrained}
-
-    alg::algType
-    fixedpoint_abstol::AType
-    fixedpoint_reltol::RType
-    fixedpoint_norm::NType
-    max_fixedpoint_iters::Int
+struct MethodOfSteps{algType,F,constrained} <: AbstractMethodOfStepsAlgorithm{constrained}
+  alg::algType
+  fpsolve::F
 end
 
 """
-    MethodOfSteps(alg; constrained::Bool=false, fixedpoint_abstol=nothing,
-                  fixedpoint_reltol=nothing, fixedpoint_norm=nothing,
-                  max_fixedpoint_iters::Int=10)
+    MethodOfSteps(alg; constrained = false, fpsolve = FPFunctional())
 
 Construct an algorithm that solves delay differential equations by the method of steps,
-where `alg` is an ODE algorithm from OrdinaryDiffEq.jl without lazy interpolation upon
-which the calculation of steps is based.
+where `alg` is an ODE algorithm from OrdinaryDiffEq.jl upon which the calculation of
+steps is based.
 
 If the algorithm is `constrained` only steps of size at most the minimal delay will be
-taken. If it is unconstrained, fixed-point iteration is applied for step sizes that exceed
-the minimal delay.
-
-The absolute and relative tolerance of the fixed-point iterations can be set by
-`fixedpoint_abstol` and `fixedpoint_reltol`, respectively, either as scalars or vectors.
-Based on these tolerances error estimates are calculated during the fixed-point iterations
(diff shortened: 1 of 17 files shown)


src/algorithms.jl
```diff
@@ -1,41 +1,21 @@
...
-Base.@pure function MethodOfSteps(alg; constrained::Bool=false, fixedpoint_abstol=nothing,
-                                  fixedpoint_reltol=nothing, fixedpoint_norm=nothing,
-                                  max_fixedpoint_iters::Int=10)
-    MethodOfSteps{typeof(alg),typeof(fixedpoint_abstol),typeof(fixedpoint_reltol),
-                  typeof(fixedpoint_norm),constrained}(alg,fixedpoint_abstol,
-                                                       fixedpoint_reltol,
-                                                       fixedpoint_norm, max_fixedpoint_iters)
-end
+MethodOfSteps(alg; constrained = false, fpsolve = FPFunctional()) =
+  MethodOfSteps{typeof(alg),typeof(fpsolve),constrained}(alg, fpsolve)
```
devmotion: I'm wondering if it is possible to get rid of the `constrained` keyword argument, and instead add a special `Constrained` solver. It feels a bit strange to specify both a solver and `constrained = false`, since in the constrained case we do not want to perform any fixed-point iterations.

src/history_function.jl
```diff
@@ -31,16 +33,21 @@ function (f::HistoryFunction)(p, t, ::Type{Val{deriv}}=Val{0}; idxs=nothing) whe
...
+  
+  tdir_solt = tdir * sol.t[end]
+  if tdir_t <= tdir_solt
     return sol.interp(t, idxs, Val{deriv}, p)
   end
 
   # history function is evaluated at time point past the final time point of
   # the current solution
-  f.isout = true
+  if tdir_t - tdir_solt > 10 * eps(tdir_t)
```
devmotion: I noticed that even for constrained time stepping due to floating point issues the solution is sometimes extrapolated. However, in that case we do not want to perform any fixed-point iterations. So basically there are two options: 1) check if the algorithm is constrained (which is a bit tricky since the DDEIntegrator does not contain any reference to the MethodOfSteps algorithm) and in that case do not check if the history function was extrapolated, or 2) generally do not perform fixed-point iterations if the extrapolation is very close to the final time point of the history and throw an error if still fixed-point iterations are performed in the constrained case. I went for the second alternative.

src/algorithms.jl
ChrisRackauckas: yeah that makes sense. Mostly it's there because the first implementation was only constrained.

src/history_function.jl
ChrisRackauckas: how would floating point issues cause accidental extrapolation?

ChrisRackauckas:
Definitely much better. However, throwing a curveball in here, would it somehow be possible to use the nlsolvers of DiffEqBase more directly? I've always wondered this but couldn't think of a good way. It would be nice to easily try out Newton iterations as well.

src/history_function.jl
devmotion: The delays are subtracted from the current time point, and thus `t - tau` can be greater than `tprev` even if `t = tprev + tau` (happens in the constrained tests, e.g.).

ChrisRackauckas: oh yes, that's terrifying. This is a good workaround.

src/algorithms.jl
devmotion: OK, in that case, should we rename the `fpsolve` argument? Or is it fine, even if `constrained` is primarily a property of the time stepping that implies that there are no fixed-point iterations?

devmotion:
>  would it somehow be possible to use the nlsolvers of DiffEqBase more directly?

Indeed, it would be great if we could reuse the implementation there. But it was not immediately clear to me how to do that (e.g., here we are actually modifying the integrator in every iteration whereas in the non-linear solvers the integrator is not updated at all), so I thought it's easier to converge from a working implementation in DelayDiffEq to something that's less separated than to start with modifying the non-linear solvers.

Issue #8, opened 2017-06-26
Anderson acceleration


ChrisRackauckas:
Instead of using Picard iterations, Anderson acceleration should be used for speed and robustness. Thankfully, NLsolve is getting some of that goodness:

https://github.com/JuliaNLSolvers/NLsolve.jl/pull/101

When that goes through, we can swap the solvers over.

ChrisRackauckas:
As addressed here: https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/12 , using NLsolve.jl is probably a bad idea for this. We should have this internal to the integrator in order to make this efficient since NLsolve.jl doesn't allow for saving the cache/re-initializing.

devmotion:
As noted here https://github.com/JuliaNLSolvers/NLsolve.jl/pull/101:

> A better implementation would use updates to qr factors, not implemented in julia yet (https://github.com/JuliaLang/julia/issues/20791). This would enable a better scaling with the history size, and more stability (through condition number monitoring).

antoine-levitt:
It would be very good to have a common codebase with NLSolve. The current code is very naive and simple, but any optimized implementation would be more complex and non-trivial. What exactly is the problem with NLSolve? If it's avoiding the allocs, I would suggest profiling to show it's a non-trivial effect first; I would think repeated allocations of the same size should be OK for the GC to handle (but I have not measured it).

ChrisRackauckas:
>What exactly is the problem with NLSolve? If it's avoiding the allocs, I would suggest profiling to show it's a non-trivial effect first; I would think repeated allocations of the same size should be OK for the GC to handle (but I have not measured it).

The allocs do matter here. Test problems use like size 10 arrays, but it's trivial for real problems to be like a system of 1,000 or 10,000 DDEs, in which case allocating that many large arrays does pretty bad. Even just reducing allocations in the Picard method when it was first created was a pretty big deal. NLsolve will need a way for us to pre-allocate for it to be viable here. NLsolve also only accepts real numbers. 

But also, this code is quite a bit more complicated to handle the complexities of the interpolation. 

https://github.com/JuliaDiffEq/DelayDiffEq.jl/blob/master/src/integrator_interface.jl#L121

Though there's probably a way to put it together with NLsolve.jl though. I would love to not double up on code, but I also don't want to take a performance and feature hit to do so.

antoine-levitt:
[...]

PR #147, opened 2019-08-12
Add Anderson acceleration

devmotion:
Fixes https://github.com/JuliaDiffEq/DelayDiffEq.jl/issues/8.

Moreover, this PR updates the type structure and dispatches of the fixed-point iteration (similar to https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/315) and uses `NLFunctional` and `NLAnderson` instead of `FPFunctional` and `FPAnderson`.

### DIFF
--- a/src/algorithms.jl
+++ b/src/algorithms.jl
@@ -7,7 +7,7 @@ struct MethodOfSteps{algType,F,constrained} <: AbstractMethodOfStepsAlgorithm{co
 end
 
 """
-    MethodOfSteps(alg; constrained = false, fpsolve = FPFunctional())
+    MethodOfSteps(alg; constrained = false, fpsolve = NLFunctional())
 
 Construct an algorithm that solves delay differential equations by the method of steps,
 where `alg` is an ODE algorithm from OrdinaryDiffEq.jl upon which the calculation of
@@ -17,5 +17,5 @@ If the algorithm is `constrained` only steps of size at most the minimal delay w
 taken. If it is unconstrained, fixed-point iteration `fpsolve` is applied for step sizes
 that exceed the minimal delay.
 """
-MethodOfSteps(alg; constrained = false, fpsolve = FPFunctional()) =
+MethodOfSteps(alg; constrained = false, fpsolve = NLFunctional()) =
   MethodOfSteps{typeof(alg),typeof(fpsolve),constrained}(alg, fpsolve)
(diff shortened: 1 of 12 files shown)


src/fpsolve/functional.jl
```diff
@@ -1,30 +1,52 @@
-function fpsolve!(integrator::DDEIntegrator, fpsolver::FPSolver{<:FPFunctionalConstantCache})
+function fpsolve!(fpsolver::FPSolver{<:Union{NLFunctional,NLAnderson},false},
```
devmotion: Should we overload `nlsolve!` instead of defining `fpsolve!`?

src/fpsolve/type.jl
```diff
@@ -1,40 +1,34 @@
...
-
-# algorithms
-struct FPFunctional{K,C} <: AbstractFPSolverAlgorithm
-  κ::K
-  fast_convergence_cutoff::C
-  max_iter::Int
+# caches
+struct FPFunctionalCache{uType,uNoUnitsType} <: DiffEqBase.AbstractNLSolverCache
+  du::uType
+  atmp::uNoUnitsType
```
devmotion: Should we also move `du` to the cache in https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/315? And add a unitless cache?

ChrisRackauckas: I think so. It's a bit weird to have something that is alg-specific but then is in the non-specific cache.

src/algorithms.jl
```diff
@@ -17,5 +17,5 @@ If the algorithm is `constrained` only steps of size at most the minimal delay w
 taken. If it is unconstrained, fixed-point iteration `fpsolve` is applied for step sizes
 that exceed the minimal delay.
 """
-MethodOfSteps(alg; constrained = false, fpsolve = FPFunctional()) =
+MethodOfSteps(alg; constrained = false, fpsolve = NLFunctional()) =
```
ChrisRackauckas: I would argue that after this settles we probably want to default to AA. But that doesn't need to happen now.

src/fpsolve/functional.jl
```diff
@@ -40,31 +62,79 @@ function fpsolve!(integrator::DDEIntegrator, fpsolver::FPSolver{<:FPFunctionalCo
...
     if η * ndu < κ && (iter > 1 || iszero(ndu))
       # fixed-point iteration converges
-      fpsolver.status = η < fpsolver.fast_convergence_cutoff ? FastConvergence : Convergence
+      fpsolver.status = η < fast_convergence_cutoff ? FastConvergence : Convergence
       fail_convergence = false
       break
     end
+
+    # perform Anderson acceleration
+    if fpcache isa FPAndersonConstantCache && iter < max_iter
```
ChrisRackauckas: I did not check this block thoroughly but it looks quite similar to the nlsolve! one so I'm assuming this chunk is good.

src/fpsolve/functional.jl
ChrisRackauckas: ehh, if these are a different cache type then I think we might as well commit to `fpsolve!`

Issue #138, opened 2019-07-29
Add other DDEFunction features


ChrisRackauckas:
Just like https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/137, Wfact and all of that can get added.

ChrisRackauckas:
Migrated to SciML/OrdinaryDiffEq.jl#3152. DelayDiffEq.jl has been migrated to OrdinaryDiffEq.jl (lib/DelayDiffEq).

PR #149, opened 2019-08-15
Add Wfact and Wfact_t support

devmotion:
This PR adds support for `Wfact` and `Wfact_t` (see https://github.com/JuliaDiffEq/DelayDiffEq.jl/issues/138).

I added more tests but probably someone should check if my implementations of `Wfact` and `Wfact_t` are correct.

The tests indicate that the support and statistics for Jacobian and W evaluations can (and should) be improved a lot in OrdinaryDiffEq. Maybe the tests are too ambitious for the current state of the implementation, but at the moment 33 of 68 tests are broken.

### DIFF
--- a/test/interface/jacobian.jl
+++ b/test/interface/jacobian.jl
@@ -8,49 +8,180 @@ using Test
     nothing
   end
 
-  function g(J, u, h, p, t)
+  njacs = Ref(0)
+  function jac(J, u, h, p, t)
+    njacs[] += 1
     J[1, 1] = 1 - h(p, t - 1)[1]
+    nothing
+  end
+
+  nWfacts = Ref(0)
+  function Wfact(W, u, h, p, dtgamma, t)
+    nWfacts[] += 1
+    W[1,1] = dtgamma * (1 - h(p, t - 1)[1]) - 1
+    nothing
+  end
+
+  nWfact_ts = Ref(0)
+  function Wfact_t(W, u, h, p, dtgamma, t)
+    nWfact_ts[] += 1
+    W[1,1] = 1 - h(p, t - 1)[1] - inv(dtgamma)
     nothing
   end
 
   h(p, t) = [0.0]
 
   # define problems
-  prob_wo_jac = DDEProblem(DDEFunction{true}(f), [1.0], h, (0.0, 40.0);
-                           constant_lags = [1])
-  prob_w_jac = DDEProblem(DDEFunction{true}(f; jac = g), [1.0], h, (0.0, 40.0);
-                          constant_lags = [1])
+  prob = DDEProblem(DDEFunction{true}(f), [1.0], h, (0.0, 40.0); constant_lags = [1])
+  prob_jac = remake(prob; f = DDEFunction{true}(f; jac = jac))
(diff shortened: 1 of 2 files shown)


test/interface/jacobian.jl
```diff
@@ -8,49 +8,180 @@ using Test
...
+    @test_broken nWfact_ts[] == sol_Wfact_t.destats.nw
 
-    @test sol_wo_jac.t ≈ sol_w_jac.t
-    @test sol_wo_jac.u ≈ sol_w_jac.u
+    # check resulting solution
+    if alg isa Rosenbrock23
+      @test sol.t ≈ sol_Wfact_t.t
+      @test sol.u ≈ sol_Wfact_t.u
+    else
+      @test_broken sol.t ≈ sol_Wfact_t.t
```
ChrisRackauckas: that means it's not working, since IIRC at this point only Rosenbrock23 is using the untransformed version still?

test/interface/jacobian.jl
```diff
@@ -8,49 +8,180 @@ using Test
...
+  function Wfact(W, u, h, p, dtgamma, t)
+    nWfacts[] += 1
+    W[1,1] = dtgamma * (1 - h(p, t - 1)[1]) - 1
+    nothing
+  end
+
+  nWfact_ts = Ref(0)
+  function Wfact_t(W, u, h, p, dtgamma, t)
+    nWfact_ts[] += 1
+    W[1,1] = 1 - h(p, t - 1)[1] - inv(dtgamma)
```
ChrisRackauckas:
```suggestion
    W[1,1] = h(p, t - 1)[1] - inv(dtgamma)
```

IIRC

devmotion: Really? I just checked the implementation in ModellingToolkit, and there it is defined as
W_t = - LinearAlgebra.I/gam + jac in
https://github.com/JuliaDiffEq/ModelingToolkit.jl/blob/fd98d1b7b82882c8abd62f9a57b9945ea035ddd2/src/systems/diffeqs/diffeqsystem.jl#L188. In our case the Jacobian is a 1x1 matrix with element `1 - h(p, t  - 1)[1]`.

ChrisRackauckas: oh yes, I did the jac wrong

test/interface/jacobian.jl
devmotion: Yeah, as I said, maybe it's a bit too ambitious at this point. The main motivation of the tests is that:
1. If we specify the Jacobian, the number of Jacobian and W evaluations should be equal to the non-zero value of our counter, and the solution should be the same.
2. If we specify Wfact or Wfact_t, the number of Jacobian evaluations should be zero and the number of W evaluations should be equal to the non-zero value of our counter, and the solution should be the same.

ChrisRackauckas: Oh, I see: the time points are equal, not that the value is correct. Yeah, I am not sure this will be the case, since symbolic CSE stuff happens on the symbolic factorization (the main source of the speedup)

devmotion:
What is this waiting for? Should the tests be changed?

PR #152, opened 2019-08-28
Fix fixed-point iteration and add proper tests

devmotion:
Embarrassingly, Anderson acceleration is broken on master since our test problem did not perform any fixed-point iterations at all. It works now, but there's some weird discrepancy between the in-place and out-of-place results (it also exists for the fixed-point iteration without Anderson acceleration, so maybe it's not only due to a possible bug in the Anderson implementation).

However, I don't want to spend more time on trying to debug this right now since once https://github.com/JuliaDiffEq/DiffEqBase.jl/pull/325 is available, we can clean everything up in a nice and well-structured way. You can have a look at a first draft in https://github.com/JuliaDiffEq/DelayDiffEq.jl/tree/fpsolver_unified - the implementation of the Anderson acceleration is gone completely and we just call into `anderson` and `anderson!` in DiffEqBase.

BTW, tests with a DDEStats struct that tracks the fixed-point iterations and convergence failures shows that with Anderson acceleration the number of function evaluations on the in-place test problem goes down from 3045 to 2385, the number of fixed-point iterations from 305 to 234, and the number of fixed-point convergence failures from 60 to 27.

### DIFF
--- a/src/fpsolve/functional.jl
+++ b/src/fpsolve/functional.jl
@@ -9,6 +9,7 @@ function fpsolve!(fpsolver::FPSolver{<:Union{NLFunctional,NLAnderson},false},
   if fpcache isa FPAndersonConstantCache
     @unpack aa_start,droptol = fpsolver.alg
     @unpack Δus,Q,R,γs = fpcache
+    local duold, uold
   end
 
   # ODE integrator caches state u at next time point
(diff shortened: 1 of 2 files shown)


devmotion:
Haha Win32 is hitting us again :joy: 

Should we just get rid of the CI tests on 32 bit? At least OrdinaryDiffEq is not tested on 32bit. But on the other hand without Win32 it would have been more difficult to notice the famous compile-time bug :thinking:

PR #153, opened 2019-08-29
Use DDEStats

devmotion:
This PR adds separate statistics for DDEs that also track the number of fixed-point iterations and convergence failures.


devmotion:
I'm worried we might observe some compile time regressions here, maybe caused by some new package versions? At least I would be surprised if just introducing a new stats type would be responsible [for the CI time outs](https://travis-ci.org/JuliaDiffEq/DelayDiffEq.jl/jobs/578148607) even though [the test ran fine yesterday](https://travis-ci.org/JuliaDiffEq/DelayDiffEq.jl/jobs/577952987).

ChrisRackauckas:
I would be surprised if this had much of an effect?

devmotion:
Yes, I agree. I'll do a bit of debugging, I don't want to merge it as long as I'm not sure why I see these performance regressions.

PR #155, opened 2019-09-07
no callback in prob

ChrisRackauckas:


Project.toml
```diff
@@ -18,7 +18,7 @@ Roots = "f2b01f46-fcfa-551c-844a-d8ac1e96c665"
 [compat]
 DataStructures = "0.17"
 DiffEqBase = "6"
-DiffEqProblemLibrary = "4.3"
+DiffEqProblemLibrary = "4"
```
devmotion: That's not useful, the tests will certainly error for any version prior to 4.3. IMO the main issue is that we test on Julia 1.0 on Travis, but all newer versions of DiffEqBiological require Julia 1.1: https://github.com/JuliaDiffEq/DiffEqBiological.jl/blob/master/Project.toml#L22

devmotion:
On Julia nightly the tests fail due to
https://github.com/JuliaDiffEq/DelayDiffEq.jl/blob/master/src/utils.jl#L136, which has to be updated.

Project.toml
devmotion: IMO we should just run the CI tests on Julia >= 1.1. Or split the DiffEqProblemLibrary into different packages for DDEs, SDEs, and so on, then we do not depend on DiffEqBiological anymore (I guess)...

ChrisRackauckas: Oh, I didn't realize Travis was running 1.0...

devmotion:
```suggestion
DiffEqProblemLibrary = "4.3"
```

PR #156, opened 2019-09-08
Add fixed-point solver statistics

devmotion:
This PR supersedes https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/153, since most commits there are already merged into master.


src/ddestats.jl
```diff
@@ -0,0 +1,34 @@
...
+  naccept::Int
+  nreject::Int
+  maxeig::Float64
+end
+
+DDEStats(x::Int = -1) = DDEStats(x, x, x, x, x, x, x, x, x, x, x, x, 0.0)
+
+function Base.show(io::IO, s::DDEStats)
+  println(io, summary(s))
+  @printf io "%-60s %-d\n" "Number of function 1 evaluations:" s.nf
```
devmotion: Should the numbers be left or right aligned?

devmotion:
The Appveyor tests fail due to the known issues on Win32...

src/ddestats.jl
ChrisRackauckas: I don't have an opinion on that.

src/ddestats.jl
```diff
@@ -0,0 +1,34 @@
...
+  nsolve::Int
+  njacs::Int
+  nnonliniter::Int
+  nnonlinconvfail::Int
+  nfpiter::Int
+  nfpconvfail::Int
+  ncondition::Int
+  naccept::Int
+  nreject::Int
+  maxeig::Float64
```
ChrisRackauckas: We should double check whether this works with AD. IIRC there is an `internalnorm` before the save here so we're good.

devmotion: It's copied from DiffEqBase: https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/master/src/destats.jl

ChrisRackauckas: okay awesome then it must work.

src/ddestats.jl
devmotion: Should DiffEqBase use `@printf` as well so the numbers are aligned?

ChrisRackauckas: Yeah it probably should. That would make it much nicer.

Issue #157, opened 2019-09-28
Test Rosenbrock methods on RADAR5_waltman


YingboMa:
Ref: https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/941

ChrisRackauckas:
@devmotion can you write a test for this? I don't know what I am looking for in this example, but we should have some form of regression test here.

devmotion:
Sure, I can add some tests, but maybe it would be useful to tag a new OrdinaryDiffEq release first?

ChrisRackauckas:
The tag should've been released already.

devmotion:
Hmm according to https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/releases the latest release is 5.17.1 which was tagged 9 days ago.

ChrisRackauckas:
Oh, I incremented Project.toml but forgot to release

PR #158, opened 2019-10-01
Add Waltman regression tests

devmotion:
Fixes https://github.com/JuliaDiffEq/DelayDiffEq.jl/issues/157.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 *.jl.cov
 *.jl.*.cov
 *.jl.mem
+Manifest.toml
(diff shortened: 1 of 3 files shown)


devmotion:
BTW, currently we test on Julia 1.1 and master - should we add Julia 1.2 (or move CI tests to Julia 1.2 to keep the load low)?

devmotion:
Test failures on Win32 are unrelated

ChrisRackauckas:
At first I thought this said "Widmann regression tests", like you're now our regression test and find all issues. It's wouldn't be false 🤣

PR #160, opened 2019-11-02
remove generic implicit methods

ChrisRackauckas:


ChrisRackauckas:
Awesome @devmotion!

ChrisRackauckas:
Since these are already undocumented since the last major, I am considering this non-breaking.

PR #162, opened 2019-11-18
Use CompatHelper to update [compat] entries

devmotion:
IMO it would be useful to figure out new releases of dependent packages in a more automated fashion to avoid issues such as https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/issues/963. Apparently https://github.com/bcbi/CompatHelper.jl (it is used, e.g., by Turing.jl) tries to solve this problem.


ChrisRackauckas:
We need to mass install CompatHelper.

PR #163, opened 2019-11-18
KenCarp4 now passes?

ChrisRackauckas:


devmotion:
Ah that's annoying. It seems our current separation into failing/passing tests does not work anymore. For Kencarp4 only https://github.com/JuliaDiffEq/DelayDiffEq.jl/blob/da60e2a8d2453a5308f588b38dab088d1b660fe2/test/integrators/sdirk.jl#L36 passes, but both other tests still fail.

PR #164, opened 2019-11-19
Avoid duplicate Travis CI tests in PRs

devmotion:
I'm not completely sure if it works, but maybe this change lets us get rid of duplicate tests in PRs and reduce the load on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 # Documentation: http://docs.travis-ci.com/user/languages/julia/
 language: julia
+branches:
+  only:
+    - master
 os:
   - linux
 #  - osx


devmotion:
I shouldn't rerun the tests so often :smile: It seems the latest changes of the handling of the DtMin return code in https://github.com/JuliaDiffEq/DiffEqBase.jl/blob/master/src/integrator_interface.jl are responsible for the test failures.

ChrisRackauckas:
ugh, how could that be? It should have only less of a chance of exiting with that condition.

devmotion:
Sure, but the tests expect the solver to exit with that condition. After the latest changes the maximum number of iterations are reached instead. So should the tests be changed? Or is there an issue that has to be fixed?

ChrisRackauckas:
I can take a look. The issue is that you won't hit that dt issue unless you really should. Why it's not the case here is odd...

devmotion:
Have you had time to take a look if there's a regression or if we should just change the tests?

ChrisRackauckas:
Awesome.

PR #170, opened 2020-01-25
Fix saveat handling

devmotion:
This PR fixes the problem in https://github.com/JuliaDiffEq/DifferentialEquations.jl/issues/556 for DDEs.


ChrisRackauckas:
Awesome. Forgot this was other places. I just put the StochasticDiffEq one in: https://github.com/JuliaDiffEq/StochasticDiffEq.jl/commit/8e6c08bd097f3b431293f05e7a1f11ccb1db634b

PR #177, opened 2020-03-21
Replace AppVeyor with Github Actions

devmotion:
Should be a lot faster. Maybe test on MacOS as well?


ChrisRackauckas:
Are there CI limits to be aware of with this?

devmotion:
They are listed here: https://help.github.com/en/actions/getting-started-with-github-actions/about-github-actions#usage-limits
I guess the most limiting factor is that it's at most 20 concurrent jobs (and 5 on MacOS). I thought it could be interesting if we want to run CI tests on Windows since I noticed that Turing switched completely from AppVeyor to Github CI and they seem to run quite fast.

ChrisRackauckas:
Sounds good enough to me to do a trial run here.

PR #181, opened 2020-05-22
Test AD without delay length estimation

devmotion:
It seems AD with ForwardDiff is broken even without delay length estimation.

### DIFF
--- a/test/interface/ad.jl
+++ b/test/interface/ad.jl
@@ -1,9 +1,13 @@
-using DelayDiffEq, Calculus, ForwardDiff
+using DelayDiffEq
+
+import FiniteDiff
+import ForwardDiff
+
 using LinearAlgebra, Test
 
 # Hutchinson's equation
 function f(du, u, h, p, t)
-  du[1] = p[1] * u[1] * (1 - h(p, t - p[2])[1] / p[3])
+  du[1] = p[2] * u[1] * (1 - h(p, t - p[1])[1] / p[3])
   nothing
 end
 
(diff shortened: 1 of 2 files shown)


test/interface/ad.jl
```diff
@@ -18,11 +22,19 @@ h(p, t) = [p[4]]
...
   p = [1.0, 1.0, 1.0, 0.0, 1.0]
-  findiff = Calculus.finite_difference(test, p)
+  findiff = FiniteDiff.finite_difference_gradient(test, p)
   fordiff = @test_logs (:warn, r"^dt <= dtmin") ForwardDiff.gradient(test, p)
-
   @test_broken findiff ≈ fordiff
+
+  # without delay length estimation
+  p = [1.0, 1.0, 0.0, 1.0]
+  findiff2 = FiniteDiff.finite_difference_gradient(p -> test(vcat(1, p)), p)
```
ChrisRackauckas, line 33: we need to do lower tolerances?

devmotion:
Test failures are due to callback failures (https://github.com/SciML/DelayDiffEq.jl/pull/181/checks?check_run_id=700558235#step:5:245):
```julia
Double callback crossing floating pointer reducer errored. Report this issue.
```
@kanav99

ChrisRackauckas:
welp add it to the list.

test/interface/ad.jl
devmotion: Lower tolerances for the DDE solver? We test with `abstol = 1e-14` and `reltol = 1e-14`.

BTW finite and forward diff yield really completely different results, for the gradient example we get
```julia
(findiff, findiff2) = ([403302.77650387725, 3.169348882494423e6, 408139.9972998477, -3.289280466135436e6, 2.8082672710581566e6], [3.169348882494423e6, 408139.9972998477, -3.289280466135436e6, 2.8082672710581566e6])
(fordiff, fordiff2) = ([0.0, -1853.468233232502, 0.0, 1275.361852383486, -2705.591974038044], [-1853.468233232502, 0.0, 1275.361852383486, -2705.591974038044])
```

ChrisRackauckas:
Interesting. Okay, yeah, gotta fix this.

kanav99:
I think i made some wrong dispatch for AD. This should be quick fix but no promises. I have witnessed such solutions before

ChrisRackauckas:
Well it's unrelated to here, so let's at least merge and do more callback fixing.

PR #184, opened 2020-06-15
Add state dependent delay citation

ChrisRackauckas:


devmotion:
I assume the test errors are caused by some of the latest changes in DiffEqBase, probably the ones in `init_call` https://github.com/SciML/DiffEqBase.jl/compare/v6.37.0...v6.38.3 (6.37.0 worked fine some days ago but 6.38.3 errors).

ChrisRackauckas:
Yes, the possibility of a logger can break inference. https://github.com/JuliaLang/julia/pull/35976 is really what's needed to do all of this properly.

devmotion:
Maybe then let's just mark all the inference tests as broken for now (and not only for a subset of algorithms)?

ChrisRackauckas:
Indeed that's my plan. What's going on is that type inference couldn't propagate static information through keyword argument splats, so anything that can change a type from a keyword argument causes an instability. So progress does it all of the time, but even before, you'd get an instability because of `save_idxs=1` changing the solution type. But since it's all in constant type information, in theory Julia could figure it out, and that PR fixes the flow of this information on Julia v1.6. So since it's not the biggest deal for performance (though it is annoying), I think it's best to just know that inference is not going to do well until v1.6, at which point everything should always infer.

ChrisRackauckas:
(and note: there's a push to have it in v1.5, and it should be discussed in the next triage meeting)

PR #189, opened 2020-10-14
fix autoswitch caching

ChrisRackauckas:
Fixes https://github.com/SciML/DifferentialEquations.jl/issues/684

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -72,6 +72,23 @@ function DiffEqBase.__init(prob::DiffEqBase.AbstractDDEProblem,
     order_discontinuity_t0 = prob.order_discontinuity_t0
   end
 
+  if alg.alg isa CompositeAlgorithm && alg.alg.choice_function isa AutoSwitch
+    auto = alg.alg.choice_function
+    alg = MethodOfSteps(CompositeAlgorithm(alg.alg.algs,
+                             OrdinaryDiffEq.AutoSwitchCache(
+                                             0,
+                                             auto.nonstiffalg,
+                                             auto.stiffalg,
+                                             auto.stiffalgfirst,
+                                             auto.maxstiffstep,
+                                             auto.maxnonstiffstep,
+                                             auto.nonstifftol,
+                                             auto.stifftol,
+                                             auto.dtfac,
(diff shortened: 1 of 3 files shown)


src/solve.jl
```diff
@@ -72,6 +72,23 @@ function DiffEqBase.__init(prob::DiffEqBase.AbstractDDEProblem,
     order_discontinuity_t0 = prob.order_discontinuity_t0
   end
 
+  if alg.alg isa CompositeAlgorithm && alg.alg.choice_function isa AutoSwitch
```
devmotion, line 75: In general, I think maybe it would be nice to reduce the number of hardcoded `if` branches and checks in the `solve` implementation (it's still a bit of a mess IMO) and use better abstractions. It is possible to move this logic to a function that creates the correct algorithm? For other non-composite algorithms it would just pass them through, I guess.

I'm also a bit worried about reassigning something (of a different type it seems?) to `alg`. Does the compiler complain here?

devmotion:
Maybe it would be good to add a test here as well?

ChrisRackauckas:
The current DelayDiffEq tests catch this. They just haven't been ran in awhile.

ChrisRackauckas:
And it was a recent change that broke this.

devmotion:
Oh I didn't know :+1:

src/solve.jl
ChrisRackauckas: While I agree with you, I don't have very much time right now and so I wanted to just put in a quick patch since it's something that hits the tutorials. I agree `solve.jl` needs to be refactored a bit, in all packages really. But I don't necessarily agree that using dispatch in this specific portion where it's a no-op for most algorithms is more readable than just doing an `if` statement: the `if` statement is really clear as to when it's hit. I think more generally algorithm modification and instantiation would get a bit of an overhaul.

>I'm also a bit worried about reassigning something (of a different type it seems?) to alg. Does the compiler complain here?

It's a v1.0 change that reassigning like this would get a different phi node and therefore the inference algorithm would treat it differently.

devmotion: Sounds good.

devmotion (approved):

PR #195, opened 2020-12-01
Re-enable CI environments

ChrisRackauckas:


.github/workflows/CI.yml
```diff
@@ -6,6 +6,40 @@ on:
...
+      - uses: julia-actions/julia-buildpkg@v1
+      - uses: julia-actions/julia-runtest@v1
+        env:
+          GROUP: ${{ matrix.group }}
+      - uses: julia-actions/julia-processcoverage@v1
+      - uses: codecov/codecov-action@v1
+        with:
+          file: lcov.info
+
+
```
devmotion, line 42: There already exists a job of name `test`. It should be removed or extended.

ChrisRackauckas: I don't know what you mean.

devmotion: The workflow contains two jobs of the same name, one for linux and one for Windows. They should be merged.

devmotion: That's also the reason for why the workflow is no valid and not even the tests on Github run.

ChrisRackauckas:
The other repos get the name in the links: is there a way to do that here?

@christopher-dG

devmotion:
There exists already a link to the Github tests since we already used it for Windows. We just have to remove the Travis badge.

ChrisRackauckas:
I mean below it says `CI / Julia 1 - ubuntu-latest - x86 (pull_request) ` instead of mentioning the environment.

devmotion:
We used a standard name for the tests before (I think the same as the PkgTemplates defaults) and did not split tests into groups. I just removed the name.

Issue #188, opened 2020-09-14
behavior of step! with tstop


slwu89:
I have a question regarding the behavior of stepping, particularly about `advance_to_tstop` , in this part of the docs: https://diffeq.sciml.ai/stable/basics/integrator/#Additional-Options
It seems that when calling `init` and passing `advance_to_tstop = true` the behavior I'd expect is that `step!` should continue stepping to that tstop.  That's not what I seem to be getting from this example code here:

```
function bc_model(du,u,h,p,t)
  p0,q0,v0,d0,p1,q1,v1,d1,d2,beta0,beta1,tau = p
  hist3 = h(p, t-tau)[3]
  du[1] = (v0/(1+beta0*(hist3^2))) * (p0 - q0)*u[1] - d0*u[1]
  du[2] = (v0/(1+beta0*(hist3^2))) * (1 - p0 + q0)*u[1] +
          (v1/(1+beta1*(hist3^2))) * (p1 - q1)*u[2] - d1*u[2]
  du[3] = (v1/(1+beta1*(hist3^2))) * (1 - p1 + q1)*u[2] - d2*u[3]
end

h(p, t) = ones(3)

tau = 1
lags = [tau]

p0 = 0.2; q0 = 0.3; v0 = 1; d0 = 5
p1 = 0.2; q1 = 0.3; v1 = 1; d1 = 1
d2 = 1; beta0 = 1; beta1 = 1
p = (p0,q0,v0,d0,p1,q1,v1,d1,d2,beta0,beta1,tau)
tspan = (0.0,100.0)
u0 = [1.0,1.0,1.0]

prob = DDEProblem(bc_model,u0,h,tspan,p; constant_lags=lags)
int = init(prob, MethodOfSteps(Tsit5());tstops = [4.5], advance_to_tstop = true, save_everystep = false)

step!(int)
int.t
```

After calling `step!` the time is 1, rather than 4.5.

devmotion:
[...]

PR #196, opened 2020-12-04
Separate propagated time stops and discontinuities

devmotion:
Fixes https://github.com/SciML/DelayDiffEq.jl/issues/188.

Requires https://github.com/SciML/OrdinaryDiffEq.jl/pull/1307.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DelayDiffEq"
 uuid = "bcd4f6db-9728-5f36-b5f7-82caef46ccdb"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>"]
-version = "5.25.3"
+version = "5.26.0"
 
 [deps]
 DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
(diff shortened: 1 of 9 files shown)


Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "DelayDiffEq"
 uuid = "bcd4f6db-9728-5f36-b5f7-82caef46ccdb"
 authors = ["Chris Rackauckas <accounts@chrisrackauckas.com>"]
-version = "5.25.2"
+version = "5.25.3"
```
ChrisRackauckas, line 4:
```suggestion
version = "5.26.0"
```

bounds change so minor.

ChrisRackauckas:
Does this have a direct test of #188 in it? I don't see what would catch that issue?

devmotion:
It's a bit hidden but https://github.com/SciML/DelayDiffEq.jl/blob/66a257a26040472a17e6617926cd4bea599cc380/test/integrators/iterator.jl#L21-L24, together with the modified `tstops = [1.5]` catches the issue. It fails on master but works with the changes in OrdinaryDiffEq and this PR.

PR #201, opened 2021-01-25
More general history function

devmotion:
Removes some OrdinaryDiffEq-specific code in the implementation of `HistoryFunction`. This allows to remvove `StochasticDelayDiffEq.HistoryFunction`.

### DIFF
--- a/src/history_function.jl
+++ b/src/history_function.jl
@@ -87,6 +87,6 @@ function (f::HistoryFunction)(val, p, t, ::Type{Val{deriv}}=Val{0}; idxs=nothing
     # handle extrapolations at initial time point
     return constant_extrapolant!(val, t, integrator, idxs, Val{deriv})
   else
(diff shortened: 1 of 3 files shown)


src/history_function.jl
```diff
@@ -87,6 +87,6 @@ function (f::HistoryFunction)(val, p, t, ::Type{Val{deriv}}=Val{0}; idxs=nothing
     # handle extrapolations at initial time point
     return constant_extrapolant!(val, t, integrator, idxs, Val{deriv})
   else
-    return OrdinaryDiffEq.current_interpolant!(val, t, integrator, idxs, Val{deriv})
+    return integrator!(val, t, Val{deriv}; idxs=idxs)
```
devmotion, line 90:
```suggestion
    return integrator(val, t, Val{deriv}; idxs=idxs)
```

ChrisRackauckas (approved):

PR #203, opened 2021-01-31
Wrap all AbstractDDEFunction which have the expected fields

wi11dey:
Since now there are more subtypes of `AbstractDDEFunction` than just `DDEFunction`, `ODEFunctionWrapper` is changed to accept any `AbstractDDEFunction` `f` that fits the interface:
- `isinplace(f)` returns `Bool`
- `f.f` is defined as the one function that contains the entire DDE.
- `f.mass_matrix` is defined
- `f.analytic` is defined
- `f.jac` is defined
- `f.jac_prototype` is defined
- `f.sparsity` is defined
- `f.Wfact` is defined
- `f.Wfact_t` is defined
- `f.paramjac` is defined
- `f.syms` is defined
- `f.colorvec` is defined

(i.e. any `AbstractDDEFunction` that "looks like" a regular `DDEFunction`)

Companion PR of https://github.com/SciML/SciMLBase.jl/pull/14.

### DIFF
--- a/src/functionwrapper.jl
+++ b/src/functionwrapper.jl
@@ -40,7 +40,7 @@ struct ODEFunctionWrapper{iip,F,H,TMM,Ta,Tt,TJ,JP,SP,TW,TWt,TPJ,S,TCV} <: DiffEq
   colorvec::TCV
 end
 
-function ODEFunctionWrapper(f::DDEFunction, h)
+function ODEFunctionWrapper(f::DiffEqBase.AbstractDDEFunction, h)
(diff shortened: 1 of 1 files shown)


ChrisRackauckas (approved):

ChrisRackauckas:
Yup great!

>AbstractDDEFunction f that fits the interface:

With the SciMLBase refactor I plan to go back through and try and document all of these interfaces. It'll take a bit, but I think it'll help the ecosystem a ton.

PR #217, opened 2021-05-25
Fix a bug and test mass matrix support

devmotion:
Recently there were some comments on Slack about problems with DDAEs so I checked and tried to solve a DDAE formulation of an SIR model. This revealed a bug (ArrayInterface was not available).

I just checked a random selection of the solvers in the DAE docs, Rosenbrock23 and Rodas4 seem to work quite well.

### DIFF
--- a/src/solve.jl
+++ b/src/solve.jl
@@ -100,8 +100,8 @@ function DiffEqBase.__init(prob::DiffEqBase.AbstractDDEProblem,
 
   progress && @logmsg(-1, progress_name, _id = _id = :DelayDiffEq, progress = 0)
 
-  isdae = prob.f.mass_matrix != I && !(typeof(prob.f.mass_matrix)<:Tuple) &&
-                                ArrayInterface.issingular(prob.f.mass_matrix)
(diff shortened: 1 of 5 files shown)


src/solve.jl
```diff
@@ -100,8 +100,8 @@ function DiffEqBase.__init(prob::DiffEqBase.AbstractDDEProblem,
 
   progress && @logmsg(-1, progress_name, _id = _id = :DelayDiffEq, progress = 0)
 
-  isdae = prob.f.mass_matrix != I && !(typeof(prob.f.mass_matrix)<:Tuple) &&
-                                ArrayInterface.issingular(prob.f.mass_matrix)
+  isdae = prob.f.mass_matrix !== I && !(prob.f.mass_matrix isa Tuple) &&
```
YingboMa, line 103: How does this fix the bug? Sorry, I just don't see the difference in logic. Is it because the equality check fails in some cases?

devmotion: The bug was that ArrayInterface was neither imported nor a dependency at all. I just reformatted the line while I was at it but there's no fix here.

PR #220, opened 2021-11-22
Add nfail to FPSolver

ChrisRackauckas:
Required for OrdinaryDiffEq 5.68 due to https://github.com/SciML/OrdinaryDiffEq.jl/pull/1488

### DIFF
--- a/src/fpsolve/type.jl
+++ b/src/fpsolve/type.jl
@@ -8,6 +8,7 @@ mutable struct FPSolver{algType,iip,uTolType,C<:DiffEqBase.AbstractNLSolverCache
   maxiters::Int
   status::OrdinaryDiffEq.NLStatus
   cache::C
+  nfail::Int
 end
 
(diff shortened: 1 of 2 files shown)


src/fpsolve/type.jl
```diff
@@ -8,6 +8,7 @@ mutable struct FPSolver{algType,iip,uTolType,C<:DiffEqBase.AbstractNLSolverCache
   maxiters::Int
   status::OrdinaryDiffEq.NLStatus
   cache::C
+  nfail::Int
```
devmotion, line 11:
```suggestion
  nfails::Int
```

devmotion (approved):
LGTM. I guess we don't have to lower bound OrdinaryDiffEq since in previous version `nfails` will just be ignored?

ChrisRackauckas:
Yup it would just be ignored so we're good.

PR #221, opened 2021-12-02
CI for LTS

ChrisRackauckas:


devmotion:
Do we want to continue testing on 32bit and both Linux and Windows? It revealed some bugs in the past but maybe it would be sufficient to test Linux and 32bit only with the latest stable release?

ChrisRackauckas:
>maybe it would be sufficient to test Linux and 32bit only with the latest stable release

I think Win32 is done, so Linux 32 is the only one needed.

PR #222, opened 2021-12-20
Update for LinearSolve.jl

ChrisRackauckas:


Project.toml
```diff
@@ -21,7 +21,7 @@ ArrayInterface = "3"
 DataStructures = "0.18"
 DiffEqBase = "6"
 NonlinearSolve = "0.3"
-OrdinaryDiffEq = "5.54"
+OrdinaryDiffEq = "6"
```
devmotion, line 24: It seems LinearSolve only affects the tests, so
```suggestion
OrdinaryDiffEq = "5.54, 6"
```
would work as well.

PR #224, opened 2021-12-22
Test master

ChrisRackauckas:


devmotion:
What upstream change causes the test errors? The exp2 PR?

ChrisRackauckas:
I believe so. I'm just slamming the computers and seeing what changed. OrdinaryDiffEq and StochasticDiffEq seem pretty fine. I know this package has some really tight bounds here and there, so a few of those had to be nudged. That's fine. This one test though, it looks like it takes a different number of steps on IIP and OOP now? That's peculiar.

ChrisRackauckas:
There's that and the manual simd ivdep on Vern9 loops. That's enough to make the floating point computation subtly different. But it still is very very very close at almost every step, so I'm going to chalk this one up as a small enough regression to slide. It's a peculiar one though.

PR #239, opened 2022-08-20
move to DDEProblemLibrary

ChrisRackauckas:

PR #240, opened 2022-08-20
Move from DiffEqProblemLibrary to DDEProblemLibrary

devmotion:
Compared with #239, this contains some additional simplifications and fixes.

### DIFF
--- a/test/Project.toml
+++ b/test/Project.toml
@@ -14,4 +14,4 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [compat]
-DDEProblemLibrary = "0.1.1"
+DDEProblemLibrary = "0.1.2"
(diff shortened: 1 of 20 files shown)


test/Project.toml
```diff
@@ -12,3 +12,6 @@ RecursiveArrayTools = "731186ca-8d62-57ce-b412-fbd966d074cd"
 SafeTestsets = "1bc83da4-3b8d-516f-aca4-4fe02f6d838f"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
+
+[compat]
+DDEProblemLibrary = "0.1.1"
```
devmotion, line 17: IMO unbounded test dependencies are quite bad. Even if upstream packages follow semver, a breaking release might break the tests. DiffEqProblemLibrary was declared to be breaking so it should not have been picked up automatically.

devmotion:
Test failures should be fixed by https://github.com/SciML/DiffEqProblemLibrary.jl/pull/92.

test/Project.toml
devmotion, line 17: Should be applied once the new version is available:
```suggestion
DDEProblemLibrary = "0.1.2"
```

PR #241, opened 2022-08-31
MassInstallAction: Install the Invalidations workflow on this repository

ranocha:
This pull request sets up the Invalidations workflow on this repository. cc: @ranocha @ChrisRackauckas

### DIFF
--- a/.github/workflows/invalidations.yml
+++ b/.github/workflows/invalidations.yml
@@ -1,0 +1,33 @@
+name: Invalidations
+on: pull_request
+
+jobs:
+  evaluate:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: julia-actions/setup-julia@v1
+      with:
+        version: '1'
+    - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/invalidations.yml
```diff
@@ -0,0 +1,33 @@
+name: Invalidations
+on: pull_request
```
devmotion, line 2: Maybe this should be run only for PRs to master as it compares invalidations with master?

devmotion: @ranocha

ranocha: I used the setup from https://github.com/SciML/MuladdMacro.jl/pull/26 and started MassInstallAction.jl, so it's too late to cancel it right now. But since someone else needs to run it again - or I need to get more rights in SciML - we can change that afterwards. Please make a PR with the change you would like to have to https://github.com/SciML/MuladdMacro.jl and we can use it as a template for the future process of adding it to every SciML repo.

PR #245, opened 2022-09-16
raise tollerance of test

oscardssmith:
Needed for https://github.com/SciML/OrdinaryDiffEq.jl/pull/1763. The previous tolerance was overly strict.

### DIFF
--- a/test/integrators/events.jl
+++ b/test/integrators/events.jl
@@ -40,8 +40,8 @@ end
     sol2 = solve(prob, alg, callback = cb, dtmax = 0.01)
     sol3 = appxtrue(sol1, sol2)
 
-    @test sol3.errors[:L2] < 1.4e-3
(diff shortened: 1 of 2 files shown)


devmotion:
So https://github.com/SciML/OrdinaryDiffEq.jl/pull/1763 introduces regressions?

oscardssmith:
It is slightly less accurate when the timespan is greater than 1, but the policy of picking an initial dt that is fixed makes a lot less sense than picking one proportional to the timespan.

devmotion:
Because of the nested structure (nesting ODE solvers inside ODE solvers) tiny regressions in OrdinaryDiffEq are amplified in DelayDiffEq. So I'm also worried about small regressions. Could one introduce a branch such that larger time spans are still handled in the same way?

oscardssmith:
This was actually nu-necessary. I only need this change for DAEs.

PR #246, opened 2022-09-16
Use both sides of tspan when determining dtmin

oscardssmith:
https://github.com/SciML/OrdinaryDiffEq.jl/pull/1762 but for DDEs.


devmotion:
@oscardssmith This PR causes a severe regression of the Waltman test problem.

PR #248, opened 2022-11-09
update symbol return codes to ReturnCode

ChrisRackauckas:


src/integrators/interface.jl
```diff
@@ -561,13 +561,13 @@ function DiffEqBase.step!(integrator::DDEIntegrator)
         if integrator.opts.advance_to_tstop
             while integrator.tdir * integrator.t < first(integrator.opts.tstops)
                 OrdinaryDiffEq.loopheader!(integrator)
-                DiffEqBase.check_error!(integrator) == :Success || return
+                DiffEqBase.check_error!(integrator) == ReturnCode.Success || return
```
devmotion, line 564: I wonder if we should change it back to `===` now that we use `ReturnCode`:

```suggestion
                DiffEqBase.check_error!(integrator) === ReturnCode.Success || return
```

ChrisRackauckas: yeah we can. I don't think it's a huge difference though since it's always going to infer it's a ReturnCode so it should just be equivalent at compile time.

PR #255, opened 2023-02-06
Fix paper build for JCON

odow:
I managed to get Whedon installed locally so I could run this, and this was the magic change that got things to work.

Here's the paper it built:

[10.21105.joss.00109.pdf](https://github.com/SciML/DelayDiffEq.jl/files/10661085/10.21105.joss.00109.pdf)

x-ref: https://github.com/JuliaCon/proceedings-review/issues/109

### DIFF
--- a/paper/paper.tex
+++ b/paper/paper.tex
@@ -15,7 +15,7 @@
 
 % configuration
 \pgfplotsset{
-    compat=1.18,
+    compat=1.16,
     % CSV files use comma as delimiter
     table/col sep=comma,
     % folders with CSV files


devmotion:
Great, thanks a lot! I was  having a hard time debugging these whedon issues - a guide for how to install it locally could be useful (if it doesn't exist yet, I think I tried to follow some instructions somewhere but it was not successful). The paper built with CI seems fine as well: https://github.com/SciML/DelayDiffEq.jl/suites/10797735417/artifacts/543009526

devmotion:
Unfortunately, Whedon still fails: https://github.com/JuliaCon/proceedings-review/issues/109#issuecomment-1418745196

PR #258, opened 2023-03-01
Replace destats with stats

avik-pal:


src/dstats.jl
```diff
@@ -1,4 +1,4 @@
-mutable struct DDEStats
+mutable struct DStats
```
devmotion, line 1: Can you explain the reasoning for renaming the struct? It's internal anyway, so the name should not matter?

devmotion: And the name `DDEStats` seems quite fitting.

devmotion: Ah maybe you just did a quick search/replace `destats` -> `stats`? In that case, can you revert these undesired changes?

avik-pal:
Locally tests pass with https://github.com/SciML/OrdinaryDiffEq.jl/pull/1889, https://github.com/SciML/DiffEqBase.jl/pull/886, and https://github.com/SciML/SciMLBase.jl/pull/391

Project.toml
```diff
@@ -24,7 +24,7 @@ DiffEqBase = "6"
 OrdinaryDiffEq = "6.33"
 RecursiveArrayTools = "2"
 Reexport = "0.2, 1.0"
-SciMLBase = "1.70"
+SciMLBase = "1.89"
```
ChrisRackauckas, line 27:
```suggestion
SciMLBase = "1.90"
```

PR #270, opened 2023-10-06
add progress_id to delaydiffeq

pepijndevos:
This is part of
https://github.com/SciML/OrdinaryDiffEq.jl/pull/2032
https://github.com/SciML/DiffEqBase.jl/pull/939
to eventually allow EnsembleProblem to log progress
https://github.com/SciML/SciMLBase.jl/pull/514


src/solve.jl
```diff
@@ -281,6 +282,7 @@ function DiffEqBase.__init(prob::DiffEqBase.AbstractDDEProblem,
                                                                                progress_steps,
                                                                                progress_name,
                                                                                progress_message,
+                                                                               progress_id,
```
ChrisRackauckas, line 285: this will need to `isdefined(fieldnames(OrdinaryDiffEq.DEOptions), :progress_id)` for backwards compatability.

pepijndevos: Is there an elegant way to just omit the argument, or you have to copy the whole instantiation?

ChrisRackauckas: I think the easiest is just to copy. It can be removed in the future.

pepijndevos: alright

src/solve.jl
```diff
@@ -241,61 +242,122 @@ function DiffEqBase.__init(prob::DiffEqBase.AbstractDDEProblem,
...
+                                                                                save_end_user,
+                                                                                callback_set,
+                                                                                isoutofdomain,
+                                                                                unstable_check,
+                                                                                verbose,
+                                                                                calck,
+                                                                                force_dtmin,
+                                                                                advance_to_tstop,
+                                                                                stop_at_next_tstop)
+
```
ChrisRackauckas, line 360:
```suggestion
end
```

PR #278, opened 2023-12-15
Fix for differential_vars update

ChrisRackauckas:


.github/workflows/CI.yml
```diff
@@ -20,7 +20,6 @@ jobs:
       matrix:
         version:
           - '1'
-          - '1.6'
```
devmotion, line 23: Could still be useful to keep testing the oldest compatible version explicitly since 1.10 will be released soon?

Project.toml
```diff
@@ -18,7 +18,7 @@ SimpleNonlinearSolve = "727e6d20-b764-4bd8-a329-72de5adea6c7"
 SimpleUnPack = "ce78b400-467f-4804-87d8-8f486da07d0a"
 
 [compat]
-ArrayInterface = "6, 7"
+ArrayInterface = "7"
 DataStructures = "0.18"
 DiffEqBase = "6.122"
 OrdinaryDiffEq = "6.49.1"
```
devmotion, line 24: Doesn't this require an OrdinaryDiffEq version bump?

Project.toml
```diff
@@ -27,4 +27,4 @@ Reexport = "0.2, 1.0"
 SciMLBase = "1.90, 2"
 SimpleNonlinearSolve = "0.1, 1"
 SimpleUnPack = "1"
-julia = "1.6"
+julia = "1.9"
```
devmotion, line 30: I guess since OrdinaryDiffEq dropped support for < 1.9?

.github/workflows/CI.yml
ChrisRackauckas: I guess. I think we'll just need to update a lot of package CIs anyways. We'll want to just bump things to v1.10 since that's the suggested LTS.

Project.toml
ChrisRackauckas: yes

.github/workflows/CI.yml
ChrisRackauckas: Yeah I'm not going to worry about this right now, but you can put a PR if you want to do that. I think when v1.10 comes out, since that's poised to be right after a huge round of a major bumps, I think we'll just bump packages to v1.10 and backport what's needed, and then hopefully that's the new LTS.

PR #289, opened 2024-04-23
ci: standardise workflows using SciML's reusable workflows

thazhemadam:
Update the workflows in this repository to use SciML's reusable workflows.
This is part of a larger effort to standardise the SciML's CI workflows for more generic and common requirements, to keep the workflows uniform and easier to maintain.


.github/workflows/Tests.yml
```diff
@@ -0,0 +1,42 @@
...
+          - Integrators
+          - Regression
+        include:
+          - coverage: false
+          - version: '1'
+            arch: x64
+            coverage: true
+        exclude: # test 32bit only with Julia 1 but not LTS
+          - version: '1.6'
+            arch: x86
```
devmotion, line 35: We don't test on Julia 1.6 anymore anyways it seems? So either 1.6 should be added above or we can remove this block completely.

thazhemadam: I wasn't sure if we wanted to test it with 1.6 or not, so I just left it in there to maintain status quo. I'll remove the block, thanks for confirming.

thazhemadam: Removed in 8770a61.

devmotion (approved):

.github/workflows/Tests.yml
```diff
@@ -0,0 +1,39 @@
...
+        group:
+          - Interface
+          - Integrators
+          - Regression
+        include:
+          - coverage: false
+          - version: '1'
+            arch: x64
+            coverage: true
+    uses: "SciML/.github/.github/workflows/tests.yml@v1"
```
ChrisRackauckas, line 33: It's not missing a - here?

thazhemadam: No, this syntax is correct, as evidenced by the workflows running as expected.

PR #291, opened 2024-07-27
Fix waltman regression test

ChrisRackauckas:
My current understanding is we hit a 1 in a gazillion chance case where the factorizations are both fine but the numerical error causes one to give a solution that goes just below zero and errors purely due to how the numerical errors align. I cannot find anything actually different between the solves other than the factorization, and as shown in the text file there, the factorizations give the same value at essentially every step to about 16 digits of accuracy. The abstol on this is 21 digits of accuracy because there are some values that shouldn't go below zero, and so it seems that this random difference in the LU actually makes a difference here, and so it's not a real broken test but really just a random fluctuation.


ChrisRackauckas:
See https://github.com/JuliaLinearAlgebra/RecursiveFactorization.jl/issues/96 for more details

PR #294, opened 2024-09-12
Fix tests not not assume Rosenbrock23 doesn't W transform

oscardssmith:
with https://github.com/SciML/OrdinaryDiffEq.jl/pull/2307/ Rosenbrock23 (like everything else) uses the W-transform so this switches the test to test Rodas5P (since otherwise there isn't a sequencing that would keep CI green).

### DIFF
--- a/test/interface/jacobian.jl
+++ b/test/interface/jacobian.jl
@@ -15,13 +15,6 @@ using Test
         nothing
     end
 
-    nWfacts = Ref(0)
-    function Wfact(W, u, h, p, dtgamma, t)
-        nWfacts[] += 1
-        W[1, 1] = dtgamma * (1 - h(p, t - 1)[1]) - 1
-        nothing
-    end
-
     nWfact_ts = Ref(0)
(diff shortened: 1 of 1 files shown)


test/interface/jacobian.jl
```diff
@@ -15,13 +15,6 @@ using Test
         nothing
     end
 
-    nWfacts = Ref(0)
-    function Wfact(W, u, h, p, dtgamma, t)
```
devmotion, line 19: Why does this PR remove the Wfact tests? We could skip them for Rosenbrock23 if needed but for TRBDF2 they should still work, shouldn't they?

oscardssmith: Wfact only gets used when transform=false and rosenbrock23 was the last solver that didn't w-transform.

oscardssmith: once I merge both of the OrdinaryDiffEq prs, I'm going to open a PR that deprecates or removes W_fact from ODEFuncton

PR #298, opened 2024-11-06
Test master again

ChrisRackauckas:


oscardssmith:
The CI failure here is very odd: Here's a reduced MWE. The things that really don't make sense to me are that the ODE version of the solver works, that swapping out the ODE solver of the DDE solver fixes it, and that when it fails, it fails on the 2nd, not the first time we resize.
```
fode(du, u, p, t) = du.=u
fdde(du, u, h, p, t) = du.=u
condition(u, t, integrator) = t-round(t)
function affect!(integrator)
    resize!(integrator, length(integrator.u) + 1)
    integrator.u[end] = 1
    nothing
end
probode = ODEProblem(fode, [0.2], (0.0, 2.0); callback = ContinuousCallback(condition, affect!))
probdde = DDEProblem(fdde, [0.2], nothing, (0.0, 2.0); callback = ContinuousCallback(condition, affect!))
solve(probode, ORK256(); dt = 0.2) #works
solve(probdde, MethodOfSteps(Tsit5()); dt = 0.2, adaptive=false) #works
solve(probdde, MethodOfSteps(Euler()); dt = 0.2) #works
solve(probdde, MethodOfSteps(ORK256()); dt = 0.2) #fails at 2nd resize
```

PR #299, opened 2024-11-06
prepare to remove OrdinaryDiffEq dependency

oscardssmith:
This does everything to remove OrdinaryDiffEq (other than actually removing it since that would probably be a slightly breaking change).


oscardssmith:
@ChrisRackauckas I think my last commit here fixes the CI issues!

ChrisRackauckas:
Except the SIR one?

ChrisRackauckas (approved):

ChrisRackauckas:
Still a major improvement, so let's merge but there's still something that has to fix the SIR thing (though I cannot recreate it locally!)

PR #304, opened 2025-01-02
Fix cache and initialization tests

jClugstor:
## Checklist

- [ ] Appropriate tests were added
- [ ] Any code changes were done in a way that does not break public API
- [ ] All documentation related to code changes were updated
- [ ] The new code follows the
  [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and
  [COLPRAC](https://github.com/SciML/COLPRAC).
- [ ] Any new documentation only uses public API
  
## Additional context
Oops, I could have sworn I made this PR last week
The cache tests that are @test_nowarn were failing because of the new ADTypes integration for OrdinaryDiffEq. Setting the chunk_size to 1 should be unecessary, since that will happen automatically anyway.

### DIFF
--- a/test/integrators/cache.jl
+++ b/test/integrators/cache.jl
@@ -48,7 +48,7 @@ println("Rosenbrock23")
 end
 
 println("Rodas4")
-@test_nowarn solve(prob, MethodOfSteps(Rodas4(chunk_size = 1)); dt = 0.5)
+@test_nowarn solve(prob, MethodOfSteps(Rodas4()); dt = 0.5)
 
 println("Rodas5")
-@test_nowarn solve(prob, MethodOfSteps(Rodas5(chunk_size = 1)); dt = 0.5)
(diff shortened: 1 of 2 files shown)


jClugstor:
This is strange... interface tests are failing for only the x86 tests. I can't really reproduce locally since I don't have an x86 machine, hmm

devmotion:
> I can't really reproduce locally since I don't have an x86 machine, hmm

IIRC you don't need a special machine, you just have to run 32bit julia (on eg a 64bit machine).

jClugstor:
Oh cool thanks, I'll check that out then.

ChrisRackauckas:
I think that's something in OrdinaryDiffEq. Should look into, but it doesn't block this.

PR #309, opened 2025-05-02
remove OrdinaryDiffEq dependency

oscardssmith:
This will be a breaking change, but will noticable reduction the precompile/load time. Unfortunately, for now at least, this still depends on `OrdinaryDiffEqDifferentiation`, `OrdinaryDiffEqNonlinearSolve`, and `OrdinaryDiffEqRosenbrock`, but it doesn't export anything from them so removing those in the future (or moving them to extensions) should be doable without breakage.


ChrisRackauckas:
It might be good to have this breaking change also have this codebase and its commit history merged into OrdinaryDiffEq as a subfolder? Then move the issues and archive the repo.

PR #310, opened 2025-05-02
switch root finder to ITP

oscardssmith:
it's faster and more robust.


ChrisRackauckas:
Need to dig into the test failures. This isn't on master?

